PR #6, opened 2016-08-16
Add xretry package with Retrier

robskillington:
cc @xichen2020 @mmihic @martin-mao @cw9

### DIFF
--- a/retry/types.go
+++ b/retry/types.go
@@ -22,7 +22,19 @@ package xretry
 
 import (
 	"time"
+
+	"github.com/m3db/m3x/errors"
 )
+
+// RetriableError returns a retriable error
+func RetriableError(err error) error {
+	return xerrors.NewRetriableError(err)
+}
+
+// NonRetriableError returns a non-retriable error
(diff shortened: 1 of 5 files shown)


retry/types.go
```diff
@@ -0,0 +1,69 @@
...
+// THE SOFTWARE.
+
+package xretry
+
+import (
+	"time"
+)
+
+// Fn is a function that can be retried
+type Fn func() error
```
xichen2020: perhaps call this `RetryableFn`?

xichen2020:
LGTM other than the nit

retry/types.go
```diff
@@ -0,0 +1,69 @@
...
+
+// ContinueFn is a function that returns whether to continue attempting an operation
+type ContinueFn func(attempt int) bool
```
mmihic: Probably need a way to indicate that you shouldn't retry on a particular error.

robskillington: Addressed

mmihic:
lgtm

PR #8, opened 2016-08-16
Add source and watch

cw9:
@robskillington @xichen2020 @mmihic


events/events.go
```diff
@@ -78,7 +78,7 @@ func (p *publisher) Get() interface{} {
 }
 
 func (p *publisher) GetAndSubscribe() (interface{}, Subscriber, error) {
-	c := make(chan struct{})
+	c := make(chan struct{}, 2)
```
xichen2020: Why does the channel need to have a capacity of 2 instead of 1?

events/events.go
```diff
@@ -0,0 +1,197 @@
...
+// Subscriber is a subscriber to an event
+type Subscriber interface {
+	Event
+
+	// Notifications returns the channel of notifications
+	Notifications() <-chan struct{}
+	// OnClose registers a function to be called on Close()
+	OnClose(func())
+	// Close closes the subscriber
+	Close()
```
xichen2020: Does the subscriber need to be responsible for closing the notification channel? Seems it'll be cleaner to let the publisher handle the close so when the notification channel is closed, the subscriber stops receiving updates, e.g.,

```
initEvent, subscriber, _ := publisher.GetAndSubscribe()  // ignore error for the sake of simplicity
for _ := range subscriber.Notifications() {
  event := subscriber.Get()
  // handle the new event
  ...
}
// notification channel closed, e.g., when publisher.Close() is called
...

File: events/events.go [hunk 3]
```diff
@@ -0,0 +1,197 @@
...
+}
+
+// NewSubscriber returns a new event subscriber
+func NewSubscriber(ch <-chan struct{}) Subscriber {
```
xichen2020: Not sure if this should be exported, it doesn't actually register it with the publisher, and if the publisher is responsible for creating subscribers, we should probably unexport this.

events/events.go
robskillington: This is so that an notification is not missed in the case of you receive a notification but before you unblock and read it another is sent and is dropped when enqueuing.

events/events.go
robskillington: Could do however it will leak if you have some long running process that wakes up, consumes and then doesn't call `Close()` at the end up receiving updates it cares about.

events/events.go
```diff
@@ -0,0 +1,197 @@
...
+type Event interface {
+	// OnGet sets the function to be executed when Get() is called
+	OnGet(GetFn)
+	// Get the latest Event
+	Get() interface{}
+}
+
+// Publisher is a publisher for an event
+type Publisher interface {
+	Event
```
robskillington: It's kind of strange that `Publisher` extends an `Event` perhaps a better name for `Event` would be `EventDataSource`?

events/events.go
```diff
@@ -0,0 +1,197 @@
...
+
+var errClosed = errors.New("publisher is closed")
+
+// GetFn returns the event
+type GetFn func() interface{}
+
+// Event is an event
+type Event interface {
+	// OnGet sets the function to be executed when Get() is called
+	OnGet(GetFn)
```
robskillington: Is this really useful?  What's it used for?

xichen2020: Feels a bit strange that we have an `OnGet` method on the Event interface. Seems this is only used to register the getter function when a subscriber is created? If so can we do something like below and get rid of the `OnGet` method?
```
func (p *publisher) GetAndSubscribe() (interface{}, Subscriber, error) {
  subscriber, _ := p.newSubscriber()  // ignore error handling here for simplicity
  return p.Get(), subscriber, nil
}

func (p *publisher) newSubscriber() (Subscriber, error) {
  p.Lock()
  if p.closed {
    p.Unlock()
    return errPublisherClosed
  }
  notificationCh := make(chan struct{}, 1)
  subscriber := newSubscriber(notificationCh, p.Get)  
  p.subscribers = append(p.subscribers, subscriber)
  p.Unlock()
  return subscriber, nil
}

type subscriber struct {
  ch <-chan struct{}
  getFn GetFn
  ...  // other fields
}

func newSubscriber(ch <-chan struct{}, getFn GetFn) Subscriber {
  return subscriber{ch: ch, getFn: getFn}
}

func (s *subscriber) Get() interface{} {
  return s.getFn()
}
```

events/events.go
```diff
@@ -0,0 +1,197 @@
...
+	return s.notificationCh
+}
+
+func (s *subscriber) OnGet(f GetFn) {
+	s.Lock()
+	s.onGetFn = f
+	s.Unlock()
+}
+
+func (s *subscriber) Get() interface{} {
```
xichen2020: How do we make sure when the publisher receives an update, the update is broadcast to all subscribers? It's unclear from the current implementation that's the case here.

events/events.go
xichen2020: I thought the channel is only used for notifying the subscribers and doesn't have the actual data. If so, it should be fine if the notification is dropped before the subscriber has a chance to read it since all the subscriber needs to know is just there's an update and it should go read it?

robskillington:
Hey so we aren't really creating pub/sub components here as you can't access in order individual events.  It's really just an implementation of the "Observer" pattern.  Also would be good to separate the things that consumers can do and producers can do so there is a "Mutable" version of each that the producer can perform special functions that the consumers can't see because they're just given a plain "Observable" interface

I'm thinking perhaps something along the lines of:

```go
type ObservableSource interface {
  Observable

  // SetAndNotify will set the observable value and notify observers
  SetAndNotify(value interface{})

  // ObserversLen will return the length of current observers
  ObserversLen() int

  // Close will close all observer notification channels
  Close()
}

type Observable interface {
  // Get returns the value
  Get() interface{}

  // GetAndObserve returns the value and an observer
  GetAndObserve() (interface{}, Observer, error)
}

type Observer interface {
  // Get returns the value of the observable
  Get() interface{}

  // C returns the channel of notifications for changes to the value of the observable
  C() <-chan struct{}

  // Close closes the observer notification channel
  Close()
}
```

mmihic:
So I was thinking we'd just create a lightweight way to watch events coming off a given source input, not a generalized mechanism for pub/sub notifications across arbitrary goroutines.  Something more like https://gist.github.com/mmihic/a43a7b60e56122be0fa7ea2f2eb5241f

cw9:
So my idea was to have separate the notifications and the data, so the publisher/notifier here does not handle data updating, it only passes the GetFn to subscribers and enable whoever creates the publisher to notify the subscribers when there is an update.

so e.g. a config service client would have it's way of talking to the config service and getting new config,  optionally it can create a publisher to allow subscribers/observers to be notified when there is an update, the notifyAll() will be called by the config client rather than by the publisher checking/polling the config client or a data source

events/events.go
cw9: Yeah now that since it's just notifications, I feel this "2" not needed

events/events.go
cw9: Sure, yeah it should not be exported

events/events.go
cw9: I was thinking about cases where some subscribers are only interested on knowing the updates but not the data, thus made the GetFn to be optional for those subscribers, but I guess they can just keep the GetFn but not call the Get() if they are not interested in the data

events/events.go
cw9: Yeah my plan was for whoever creates the publisher to trigger the notification rather than the publisher to watch for updates.

cw9:
Looking at the interface proposed by Rob and Matt's snippet, I feel the main difference is whether we want a SetAndNotify() or WatchAndNotify() or just Notify()

cw9:
@mmihic is there any preference between having the source polling value updates to trigger notifications or having the creator of the source update the value in the source to trigger the notifications? I feel the later one is more light-weight because the former one requires the user to create a poll() function?

cw9:
I kinda want to do something like:

```
type Observer interface {
	xclose.Closer

	C() <-chan struct{}
	Get() interface{}
}

type Observable interface {
	xclose.Closer

	Get() interface{}
	Subscribe() (Observer, error)
	SetAndNotify(interface{}) error
}
```

And then the Source does not inherit Observable (because notify should not be called explicitly) but uses the Observable to deal with notifications like:

```
type Source interface {
	xclose.Closer

	GetAndSubscribe() (interface{}, Observer, error)
}

type SourceInput interface {
	Poll() (interface{}, error)
}

type source struct {
	sync.RWMutex

	input SourceInput
	o     Observable
}

```

xichen2020:
Hey yeah makes sense, although I'll still keep `Get()` in the `Source` interface, which is useful if say I only want to get the latest value and don't care about updates after that. Also I'd rename `Source` to `ObservableSource` just to be more self explanatory.

Alternatively, you can merge the `Observable` interface with `Source` interface into `ObservableSource` interface and make `SetAndNotify` unexported. Keeping them separate is slightly more reusable, whereas merging them makes the interfaces a bit simpler. I have no strong preference towards either.

observe/obervable.go
```diff
@@ -0,0 +1,164 @@
+package observe
+
+import (
+	"errors"
+	"sync"
+
+	"github.com/m3db/m3x/close"
+)
+
+var ErrClosed = errors.New("closed")
```
xichen2020: Does this error need to be exported?

observe/obervable.go
```diff
@@ -0,0 +1,164 @@
...
+}
+
+// Observable can be observed by Observers
+type Observable interface {
+	xclose.SimpleCloser
+
+	// Get returns the latest value
+	Get() interface{}
+	// Subscribe returns an Observer that will be notified on updates
+	Subscribe() (Observer, error)
```
xichen2020: Might be better to do `GetAndSubscribe` instead to return the initial value along with the observer so the observer doesn't need to call `Get` to get it

observe/obervable.go
```diff
@@ -0,0 +1,164 @@
...
+	}
+}
+
+type observer struct {
+	sync.Mutex
+
+	o       Observable
+	c       <-chan struct{}
+	closed  bool
+	closeFn func()
```
xichen2020: Perhaps typedef `func()` as something like `closer` to make it clear?

observe/obervable.go
```diff
@@ -0,0 +1,164 @@
...
+
+	if o.closed {
+		return
+	}
+
+	o.closed = true
+
+	if o.closeFn != nil {
+		onCloseFn := o.closeFn
+		go onCloseFn()
```
xichen2020: Why do we need to spin up a go routine here?

observe/source.go
```diff
@@ -0,0 +1,71 @@
...
+}
+
+func (s *source) Close() {
+	s.Lock()
+	defer s.Unlock()
+	if s.closed {
+		return
+	}
+	s.closed = true
+	go s.o.Close()
```
xichen2020: Do we need to spin up a goroutine here?

observe/obervable.go
cw9: yeah nope, will change

observe/obervable.go
cw9: sure, can change

observe/obervable.go
cw9: will do

observe/obervable.go
cw9: we don't have to, but using a go routine can enable the observer.Close() to not wait on the clean ups in observable. I feel that for the observer itself, setting closed = true is good enough. I can do either way.

observe/source.go
cw9: we don't have to, but we don't need to wait on s.o.Close(), I'm ok either way.

observe/obervable.go
xichen2020: I'm in favor of getting rid of the goroutine. Seems like the only way to get blocked here is to acquire the lock, which I'm not too concerned about.

observe/source.go
xichen2020: I'm in favor of getting rid of the goroutine. Seems like the only way to get blocked here is to acquire the lock, which I'm not too concerned about.

mmihic:
macro comment (which may sound like a nit): the naming terminology here is rather verbose and Java-like.  I'd ask that we switch the terminology to be more concise and Go-like:  e.g. Just ``Source`` (not ``ObservableSource`` - the qualifier is derived from the package name already), ``Watch`` rather than ``GetAndSubscribe``, ``Watch`` rather than ``Observable``, ``Update`` rather than ``SetAndNotify``, etc.

observe/observable_test.go
```diff
@@ -0,0 +1,161 @@
...
+	valueMap := make(map[int]int, subLen)
+	for i := 0; i < subLen; i++ {
+		_, s, err := p.Observe()
+		assert.NoError(t, err)
+		subMap[i] = s
+		valueMap[i] = -1
+	}
+
+	for i := 0; i < subLen; i++ {
+		i := i
```
xichen2020: any reason we need to create a local `i` here?

cw9: nope, forgot to remove this.

cw9:
Question: do we want to guarantee observers get initialized/non-nil value from the source?

xichen2020:
Either way should be fine, although I'm slightly leaning towards making `Watch` nonblocking and the watchers receive notifications in a goroutine. If the blocking semantics is necessary, can always add a `blockingTillInitialized` flag to the interface or a separate `WatchSync` method.

robskillington:
Yeah let's make it agnostic then producers of data who return the `Watch` can work out if they want to block to the upstream caller before returning the watch or not.  i.e. they can implement both patterns if they wish.

observe/obervable.go
```diff
@@ -0,0 +1,167 @@
...
+
+		if o.closed {
+			return
+		}
+
+		close(c)
+
+		for i, s := range o.active {
+			if s == c {
+				o.active = append(o.active[:i], o.active[i+1:]...)
```
robskillington: Perhaps break after this statement.

observe/obervable.go
```diff
@@ -0,0 +1,167 @@
...
+		o.Lock()
+		defer o.Unlock()
+
+		if o.closed {
+			return
+		}
+
+		close(c)
+
+		for i, s := range o.active {
```
robskillington: Perhaps use `for i := 0; i < len(o.active); i++` just because you're modifying `o.active` inside the loop?  i.e. just to avoid mutations inside a foreach construct.

observe/obervable.go
cw9: done

observe/obervable.go
cw9: done

watch/source.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+func (s *source) run() {
+	for !s.isClosed() {
+		data, err := s.input.Poll()
+		if err != nil {
+			s.logger.Errorf("error polling input source: %v", err)
+			continue
+		}
+		s.o.Update(data)
```
robskillington: Probably should not update if closed during the poll call?

watch/watch.go
```diff
@@ -0,0 +1,159 @@
...
+
+	"github.com/m3db/m3x/close"
+)
+
+var errClosed = errors.New("closed")
+
+type closer func()
+
+// Watch watches a Watchable instance, can get notification when the Watchable updates
+type Watch interface {
```
robskillington: We going to make this a `Watcher` then?

watch/source.go
cw9: yeah I can make the extra check here, but the s.o.Update() will do a check on closed too and will not do work if closed.

watch/source.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	go s.run()
+	return s
+}
+
+type source struct {
+	sync.RWMutex
+
+	input  SourceInput
```
cw9: will do

watch/watch.go
cw9: discussed offline, gonna stick with Watch

robskillington:
LGTM

PR #10, opened 2016-08-18
Add a CapturingMatcher, for capturing mock args...

mmihic:
...for deeper introspection

R: @xichen2020 @robskillington

### DIFF
--- a/mock/capturing_matcher.go
+++ b/mock/capturing_matcher.go
@@ -1,0 +1,36 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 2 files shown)


mock/capturing_matcher.go
```diff
@@ -0,0 +1,36 @@
...
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package xmock
+
+// A CapturingMatcher is a gomock.Matcher that "captures" the argument for
+// deeper introspection after the call chain completes
+type CapturingMatcher struct {
```
xichen2020: Just curious, how is the `CapturingMatcher` going to be used? It doesn't look like the auto-generated mocks take in custom matchers, so I assume this is for hand-crafted gomock mocks?

mmihic: autogenerated mocks take custom matchers.  cf https://github.com/m3db/m3cluster/blob/mmihic-changelists/changeset/manager_test.go#L49

xichen2020: I see so it's passed in to methods in place of the actual arg and the captured value is used for assertions later, yeah useful for mocking out methods that take pointers etc. Sounds good.

xichen2020:
LGTM

robskillington:
You can also use the `.Do(func(arg1 arg1type, arg2 arg2type) { capturedArg1 = arg1 })` to capture the argument instead of a matcher btw.  In fact that might be cleaner than creating a matcher that just captures the argument as a matcher is orthogonal to capturing data from the call.

kobolog:
This PR is pretty much stale, are we going to merge it or should we just abandon it? Let's not let PRs to hang around w/o any activity for more than a few weeks.

PR #12, opened 2016-08-23
Add WaitInit function to Source

cw9:
Add a Initialized function to expose the status of the source, so the users can wait on it to be initialized if needed.

### DIFF
--- a/watch/source.go
+++ b/watch/source.go
@@ -86,11 +86,9 @@ func (s *source) run() {
 
 		err = s.w.Update(data)
 
-		if err == nil && !s.isInitialized() {
-			s.Lock()
+		if err == nil && !s.initialized {
 			close(s.ch)
 			s.initialized = true
-			s.Unlock()
 		}
 	}
 }
@@ -101,15 +99,7 @@ func (s *source) isClosed() bool {
 	return s.closed
 }
 
-func (s *source) isInitialized() bool {
-	s.RLock()
-	defer s.RUnlock()
-	return s.initialized
-}
-
 func (s *source) Initialized() <-chan struct{} {
-	s.RLock()
-	defer s.RUnlock()
 	return s.ch
 }
 
--- a/watch/source.go
+++ b/watch/source.go
@@ -41,8 +41,8 @@ type SourceInput interface {
 type Source interface {
 	xclose.SimpleCloser
 
-	// Initialized returns a channel that blocks until the Source was initialized
-	Initialized() <-chan struct{}
+	// WaitInit returns a channel that blocks until the Source was initialized
+	WaitInit() <-chan struct{}
 	// Watch returns the value and an Watch
 	Watch() (interface{}, Watch, error)
 }
@@ -53,7 +53,7 @@ func NewSource(input SourceInput, logger xlog.Logger) Source {
 		input:  input,
 		w:      NewWatchable(),
 		logger: logger,
-		ch:     make(chan struct{}),
+		initCh:     make(chan struct{}),
 	}
 
 	go s.run()
@@ -67,7 +67,7 @@ type source struct {
 	w           Watchable
 	closed      bool
 	initialized bool
-	ch          chan struct{}
+	initCh      chan struct{}
 	logger      xlog.Logger
 }
 
@@ -87,7 +87,7 @@ func (s *source) run() {
 		err = s.w.Update(data)
 
 		if err == nil && !s.initialized {
-			close(s.ch)
+			close(s.initCh)
 			s.initialized = true
 		}
 	}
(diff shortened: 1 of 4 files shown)


watch/source.go
```diff
@@ -21,19 +21,24 @@
...
+
 // SourcePollFn provides source data
 type SourcePollFn func() (interface{}, error)
 
 // Source polls data by calling SourcePollFn and notifies its watches on updates
 type Source interface {
 	xclose.SimpleCloser
 
+	// Initialized returns whether the Source was initialized
+	Initialized() bool
```
mmihic: Shouldn't this be a channel, so that callers can wait on it?

cw9: I'm thinking having the caller to figure out a way to wait on it, so you mean return a channel that'll be closed when initializing?

mmihic: Correct.

cw9: Sure, will do

watch/source.go
```diff
@@ -21,19 +21,24 @@
...
 
 import (
+	"errors"
 	"sync"
 
 	"github.com/m3db/m3x/close"
 	"github.com/m3db/m3x/log"
 )
 
+var ErrSourceClosed = errors.New("source closed")
```
xichen2020: Do we need to export this?

watch/source.go
```diff
@@ -53,20 +58,33 @@ func NewSource(poll SourcePollFn, logger xlog.Logger) Source {
...
+	w           Watchable
+	closed      bool
+	initialized bool
+	logger      xlog.Logger
 }
 
 func (s *source) run() {
 	for !s.isClosed() {
 		data, err := s.poll()
+		if err == ErrSourceClosed {
```
xichen2020: Which method returns this error?

watch/source.go
```diff
@@ -53,20 +58,33 @@ func NewSource(poll SourcePollFn, logger xlog.Logger) Source {
...
+			s.logger.Errorf("Upstream source is closed")
+			s.Close()
+			return
+		}
 		if err != nil {
```
xichen2020: Should there be a polling period here and possibly with backoff/jitter when a polling error occurs?

cw9: The backoff/jitter could be left in the impl of the pollFn, but probably safer to guard it here too, I'm ok either way.

watch/source.go
cw9: This error is expected from the SourcePollFn, the impl of SourcePollFn could return this error to tell Source to stop polling it.

watch/source.go
```diff
@@ -53,20 +58,33 @@ func NewSource(poll SourcePollFn, logger xlog.Logger) Source {
 type source struct {
 	sync.RWMutex
 
-	poll   SourcePollFn
-	w      Watchable
-	logger xlog.Logger
-	closed bool
+	poll        SourcePollFn
```
mmihic: Just say this is a function pointer - we should really make the poller be a interface, so it could hold state about the last poll.

watch/source.go
mmihic: Another reason to make the SourcePoller an interface not just a function.

watch/source.go
cw9: could we just wrap the SourcePollFn with a retrier? what other benefits do we get from an interface?

watch/source.go
```diff
@@ -55,18 +62,33 @@ type source struct {
 
 	poll   SourcePollFn
 	w      Watchable
-	logger xlog.Logger
 	closed bool
+	i      bool
```
xichen2020: call this `initialized` perhaps?

cw9: it was conflicting with initialized(), renamed both

watch/source.go
cw9: changed to interface

watch/source.go
```diff
@@ -18,32 +18,42 @@
...
+	// Poll will be called by Source for data, any backoff/jitter logic should be handled here
+	Poll() (interface{}, error)
+}
 
 // Source polls data by calling SourcePollFn and notifies its watches on updates
 type Source interface {
 	xclose.SimpleCloser
 
+	// Initialized returns a channel that blocks until the Source was initialized
+	Initialized() <-chan struct{}
```
robskillington: Perhaps a better name is `WaitInitialized`?

watch/source.go
```diff
@@ -18,32 +18,42 @@
...
 
 // NewSource returns a Source
-func NewSource(poll SourcePollFn, logger xlog.Logger) Source {
+func NewSource(input SourceInput, logger xlog.Logger) Source {
 	s := &source{
-		poll:   poll,
+		input:  input,
 		w:      NewWatchable(),
 		logger: logger,
+		ch:     make(chan struct{}),
```
robskillington: Perhaps `initCh` is better?  `ch` makes it sound like it actually might back the data or something.

watch/source.go
xichen2020: or `WaitUntilInitialized`

robskillington:
LGTM

watch/source.go
cw9: Renamed to WaitInit

PR #14, opened 2016-08-24
Removing metrics

jra3:
We're moving to github.com/uber-go/tally


xichen2020:
LGTM

We'll need to fix up the dependencies in other m3db repos afterwards.

PR #16, opened 2016-09-01
Updating watchable functionality for new watches

martin-mao:
Updating the watchable functionality for creating watches - giving an initial update to the created watch if the watchable has a value.

### DIFF
--- a/watch/watch.go
+++ b/watch/watch.go
@@ -82,6 +82,10 @@ func (w *watchable) Watch() (interface{}, Watch, error) {
 	}
 
 	c := make(chan struct{}, 1)
+	if w.value != nil {
+		c <- struct{}{}
+	}
+
 	w.active = append(w.active, c)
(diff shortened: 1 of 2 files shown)


cw9:
LGTM

watch/watch.go
```diff
@@ -82,6 +82,10 @@ func (w *watchable) Watch() (interface{}, Watch, error) {
 	}
 
 	c := make(chan struct{}, 1)
+	if w.value != nil {
```
mmihic: nit: can we do this outside of the write lock?  

````golang
notify := w.value != nil
w.active = append(w.active, c)
closeFn := w.closeFunc(c)
w.Unlock()

if notify {
   select {
      case c <- struct{}{}:
      default:
   }
}
````

That way we minimize the amount of done with the lock held and reduce contention.

PR #17, opened 2016-09-01
Moving operations out of watchable lock

martin-mao:


watch/watch.go
```diff
@@ -82,14 +82,18 @@ func (w *watchable) Watch() (interface{}, Watch, error) {
...
-		c <- struct{}{}
-	}
-
+	notify := w.value != nil
 	w.active = append(w.active, c)
-	closeFn := w.closeFunc(c)
 	w.Unlock()
 
+	if notify {
+		select {
```
cw9: this select is unnecessary? sending on this channel should be non-blocking since it's just created?

martin-mao: Yeah we need the select - It's already been added to all the slice of ch's, so technically it's possible that an update is registered to it after we unlock, but before we push to it here.

mmihic:
lgtm

PR #24, opened 2016-10-02
Add debug RWMutex

robskillington:
cc @xichen2020 @martin-mao @cw9 @mmihic


debug/mutex.go
```diff
@@ -0,0 +1,164 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
martin-mao: Is this mostly copied from the Go sync Lib? If large sections are, do we need to add their copyright here?

robskillington: It's actually purely new code that is simply a wrapper around `sync.RWMutex`

PR #26, opened 2016-11-23
Move some common stuff over from M3DB.

kobolog:
This PR moves `pool` and `instrument` common packages from M3DB to M3X since they are going to be reused in other projects.


kobolog:
Yes, of course, but I'd rather do it in a separate PR since this one introduces no code changes thus requires no proof of correctness for now.

kobolog:
Yes - https://github.com/m3db/m3db/pull/167.

PR #34, opened 2016-12-26
Add utility function to start server-side accept loops

xichen2020:
cc @robskillington @martin-mao @kobolog @cw9 

This PR adds `StartAcceptLoop` function to start server-side accept loops, returning connections channel and error channel for handlers to consume.

### DIFF
--- a/net/server.go
+++ b/net/server.go
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package net
+package xnet
 
 import (
 	"net"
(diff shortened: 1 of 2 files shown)


net/server.go
```diff
@@ -0,0 +1,68 @@
...
+// temporary network errors. Fatal errors are returned via the
+// error channel with the listener closed on return
+func StartAcceptLoop(l net.Listener, retrier xretry.Retrier) (<-chan net.Conn, <-chan error) {
+	var (
+		connCh = make(chan net.Conn)
+		errCh  = make(chan error)
+	)
+
+	go func() {
+		defer l.Close()
```
cw9: nit: seems l.Close() can only happen after breaking the for loop? can move l.Close() after the for loop?

xichen2020: Correct. I feel it's slightly more readable to use a deferred close. In practice it's only closed when the connection is broken and not on the critical path.

PR #35, opened 2016-12-27
Add checked package and checked object & bytes pool

robskillington:
cc @kobolog @xichen2020 @cw9 @martin-mao


robskillington:
Needs some further tests for coverage but would like feedback on the API, etc.

pool/checked_bytes.go
```diff
@@ -0,0 +1,103 @@
...
+	if bytes.Len() == bytes.Cap() {
+		newBytes := pool.Get(bytes.Cap() * 2)
+
+		// Inc the ref to read/write to it
+		newBytes.IncRef()
+		newBytes.Resize(bytes.Len())
+
+		copy(newBytes.Get(), bytes.Get())
+
+		bytes = newBytes
```
xichen2020: Why not simply `DecRef` here instead of using a separate `swapped` bool to keep track?

checked/types.go
```diff
@@ -0,0 +1,100 @@
...
+
+// Finalize will call the function literal as a finalizer.
+func (fn FinalizerFn) Finalize() {
+	fn()
+}
+
+// Ref is an entity that checks ref counts.
+type Ref interface {
+	// IncRef increments the ref count to this entity.
+	IncRef()
```
xichen2020: I'm not entirely sure how the APIs are supposed to be used. Is it something along the lines of:

* When someone gets a new reference to the refcounted object, `IncRef` must be called
* When someone loses reference to an existing refcounted object, `DecRef` must be called
* When someone reads/writes a refcounted object, `IncRead/DecRead/IncRef/DecRef` must be called?

checked/debug.go
```diff
@@ -0,0 +1,201 @@
...
+
+	return debugger
+}
+
+func tracebackEvent(c *RefCount, ref int, e debuggerEvent) {
+	d := getDebuggerRef(c)
+	stack := debug.Stack()
+	endlines := 0
+	skipPrologue := 1
+	skipEntry := 2
```
xichen2020: Why are we skipping two entries here?

checked/types.go
robskillington: Correct.  Then when you want the finalizer to run, e.g. to return something to a pool, you call the `thing.Finalize()` and it will run its finalizer and make sure the ref count is zero.

checked/types.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// DecRef decrements the ref count to this entity.
+	DecRef()
+
+	// XfrRef transfers the ref to this entity from object to another.
+	XfrRef()
```
robskillington: Sure thing, sounds good.

checked/types.go
```diff
@@ -0,0 +1,100 @@
...
+	Ref
+
+	// IncReads increments the reads count to this entity.
+	IncReads()
+
+	// DecReads decrements the reads count to this entity.
+	DecReads()
+
+	// NumReads returns the active reads count to this entity.
+	NumReads() int
```
robskillington: I did have that before, I'll return it to that.

checked/types.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+// ReadWrite is an entity that checks reads and writes.
+type ReadWrite interface {
+	Read
+
+	// IncWrites increments the writes count to this entity.
```
robskillington: Not sure if Write makes sense without Read?

checked/types.go
```diff
@@ -0,0 +1,100 @@
...
+	IncWrites()
+
+	// DecWrites decrements the writes count to this entity.
+	DecWrites()
+
+	// NumWrites returns the active writes count to this entity.
+	NumWrites() int
```
robskillington: I did have that before, I'll return it to that.

pool/checked_bytes.go
robskillington: Because some people may want to keep reference to the old bytes if it's swapped.  Also it's bad practice to decrement the ref on behalf of the caller.  The caller should determine ownership/its own ref management.

I had it this way before and it was very messy.

checked/debug.go
robskillington: Because they're always just the "traceback..." functions which is just noise.

checked/types.go
robskillington: Sure, can make it composition.

checked/types.go
xichen2020: I see, this feels intrusive though because developers would need to remember to follow this workflow mentally when switching from the unchecked pools to checked pools, or else their code would panic in unexpected places. Ideally we could make the APIs more friendly to developers. Do you think that'll be achievable, or are we going to chalk this up to the overhead of using checked pools?

pool/types.go
```diff
@@ -20,14 +20,20 @@
...
 package pool
 
-import "github.com/m3db/m3x/instrument"
+import (
+	"github.com/m3db/m3x/checked"
+	"github.com/m3db/m3x/instrument"
+)
 
-// Allocator allocates an object for a pool.
+// Allocator allocates an object for a pool
```
robskillington: Sure, will add back.  Unfortunately this is very inconsistently applied over all our projects.

checked/types.go
xichen2020: Possibly, e.g., a byte buffer could be treated as write-only in one context and read write in another, in which cases it'll be accessed using `Write` and `ReadWrite` interfaces

pool/checked_bytes.go
xichen2020: Ideally the caller should be unaware of whether the bytes have been swapped or not since that can be treated as an implementation detail. Also I'd consider decrementing the ref here as managing the ownership of newBytes rather than the bytes passed in.

Anyway I don't feel strongly about this, if it's messy to implement, it's fine to leave it as is.

checked/debug.go
xichen2020: Sounds good

pool/types.go
xichen2020: Yes, rather unfortunate :\

xichen2020:
Also would be good to bring the coverage back to similar levels as before, 6.2% is a rather significant drop

checked/types.go
robskillington: I attempted quite a few different APIs, the simplest was just exposing these as is.  There could be some syntactic sugar for inc'ing multiple things, i.e.:
```go
checked.IncRef(a, b, c)
defer checked.DecRef(a, b, c)
```

But at the end of the day each element is usually very much one at a time handled so its not really used all that necessarily and I'm wary to add it into the mix as broadening the API might make it too hard for first timers to grasp.

pool/checked_bytes.go
robskillington: I did this a few different ways, and being more explicit about the ref management made it much simpler and pragmatic.  

Otherwise it became too magic and it was very hard to reason about the code inside loops, etc.

robskillington:
Yeah working on the coverage.

robskillington:
@xichen2020 how's that for coverage, +3.3%! 😀

xichen2020:
Noice!

checked/debug.go
```diff
@@ -0,0 +1,206 @@
...
+		c.finalizer = &debuggerRef{finalizer: c.finalizer}
+		return c.finalizer.(*debuggerRef)
+	}
+
+	return debugger
+}
+
+func tracebackEvent(c *RefCount, ref int, e debuggerEvent) {
+	d := getDebuggerRef(c)
+	stack := debug.Stack()
```
kobolog: I think it's quite inefficient in terms of space to keep actual stack traces for events. Originally, I've been storing PC values obtained via `runtime.Callers()`, which are simply `uintpts`s and formatting them into an actual stack trace only when requested. I don't think we can actually run M3DB with a reasonable retention period, keep stringified stack traces for each traced variable and not run out of memory at the same time.

robskillington: Sure thing, we can store program counters and convert to stack on request.  Sounds like it will cost a lot less in terms of memory. 

We'll only be storing this data if `SetTraceback(true)` is set which by default will be off.

PR #38, opened 2017-01-04
Use program counters instead of stacks to track ref count usage

robskillington:
cc @kobolog @xichen2020 @cw9 @martin-mao

### DIFF
--- a/checked/debug.go
+++ b/checked/debug.go
@@ -23,19 +23,28 @@ package checked
 import (
 	"bytes"
 	"fmt"
-	"runtime/debug"
+	"runtime"
+	"sync"
 	"time"
 )
 
 const (
-	defaultTraceback       = false
-	defaultTracebackCycles = 3
+	defaultTraceback         = false
+	defaultTracebackCycles   = 3
+	defaultTracebackMaxDepth = 64
 )
 
 var (
-	traceback       = defaultTraceback
-	tracebackCycles = defaultTracebackCycles
-	panicFn         = defaultPanic
+	traceback            = defaultTraceback
+	tracebackCycles      = defaultTracebackCycles
+	tracebackMaxDepth    = defaultTracebackMaxDepth
+	tracebackCallersPool = sync.Pool{New: func() interface{} {
+		return make([]uintptr, tracebackMaxDepth)
+	}}
+	tracebackEntryPool = sync.Pool{New: func() interface{} {
+		return &debuggerEntry{}
+	}}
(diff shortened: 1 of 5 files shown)


checked/debug.go
```diff
@@ -114,24 +128,31 @@ type debugger struct {
...
+	entry.ref = ref
+	entry.pc = pc
+	entry.t = time.Now()
+	d.entries[idx] = append(d.entries[idx], entry)
 	if event == decRefEvent && ref == 0 {
 		if len(d.entries) == tracebackCycles {
 			// Shift all tracebacks back one if at end of traceback cycles
+			slice := d.entries[0]
+			for i, entry := range slice {
+				tracebackCallersPool.Put(entry.pc)
```
kobolog: I wonder what this code would actually do? `entry.pc` is a `[]uintptr`, i.e. it's a struct with a pointer to a vector on the heap. When we put it back into the pool, it gets casted into an `interface{}`, i.e. now it's a pointer to a struct with a pointer to a vector on the heap, and it ends up being stored somewhere in the pool. Then we nil it, effectively setting the pointer in the struct to nil and releasing the vector. Does it mean that inside the pool we now have a struct with a nil pointer, thus not really pooling the actual heap memory but only the slice headers?

checked/debug.go
```diff
@@ -191,16 +229,13 @@ func getDebuggerRef(c *RefCount) *debuggerRef {
 
 func tracebackEvent(c *RefCount, ref int, e debuggerEvent) {
 	d := getDebuggerRef(c)
-	stack := debug.Stack()
-	endlines := 0
-	skipPrologue := 1
+	depth := tracebackMaxDepth
+	pc := tracebackCallersPool.Get().([]uintptr)
+	if capacity := cap(pc); capacity < depth {
```
kobolog: How would we arrive into this branch?

checked/debug.go
robskillington: I believe that the `interface{}` version of the `[]uintptr` retains the pointer to the still allocated array.  

Nil'ing it out below only releases the root from the `*debuggerEntry` to the `[]uintptr` which the pool still holds onto.

checked/debug.go
robskillington: We could potentially reach this case if someone changes the traceback cycles during the runtime of the program.  Essentially just some defensive programming.

checked/debug.go
kobolog: Yep, just verified this behavior.

PR #40, opened 2017-01-17
Add leak detection support.

kobolog:
This PR adds support for leak detection tracking in checked objects by utilizing Go runtime finalizers. Essentially, each checked object can be marked for tracking by calling `Ref::TrackObject(v interface{})`, which will calculate the tracked allocation size and set a finalizer on the object that will determine whether it was garbage-collected while still having references. All detected leaks could then be queried with `DumpLeaks()` function.

### DIFF
--- a/checked/bytes.go
+++ b/checked/bytes.go
@@ -54,6 +54,7 @@ func NewBytes(value []byte, opts BytesOptions) Bytes {
 		value: value,
 	}
 	b.SetFinalizer(b)
+	b.TrackObject(b.value)
 	return b
 }
 
--- a/checked/ref.go
+++ b/checked/ref.go
@@ -39,18 +39,13 @@ type RefCount struct {
 // IncRef increments the reference count to this entity.
 func (c *RefCount) IncRef() {
 	n := atomic.AddInt32(&c.ref, 1)
-	if traceback {
-		tracebackEvent(c, int(n), incRefEvent)
-	}
+	tracebackEvent(c, int(n), incRefEvent)
 }
 
 // DecRef decrements the reference count to this entity.
 func (c *RefCount) DecRef() {
 	n := atomic.AddInt32(&c.ref, -1)
-
-	if traceback {
-		tracebackEvent(c, int(n), decRefEvent)
-	}
+	tracebackEvent(c, int(n), decRefEvent)
 
 	if n < 0 {
 		err := fmt.Errorf("negative ref count, ref=%d", n)
@@ -60,9 +55,7 @@ func (c *RefCount) DecRef() {
 
 // MoveRef signals a move of the ref to this entity.
 func (c *RefCount) MoveRef() {
-	if traceback {
-		tracebackEvent(c, c.NumRef(), moveRefEvent)
-	}
+	tracebackEvent(c, c.NumRef(), moveRefEvent)
 }
 
 // NumRef returns the reference count to this entity.
@@ -73,20 +66,15 @@ func (c *RefCount) NumRef() int {
 // Finalize will call the finalizer if any, ref count must be zero.
 func (c *RefCount) Finalize() {
 	n := c.NumRef()
-
-	if traceback {
-		tracebackEvent(c, int(n), finalizeEvent)
-	}
+	tracebackEvent(c, int(n), finalizeEvent)
 
 	if n != 0 {
 		err := fmt.Errorf("finalize before zero ref count, ref=%d", n)
 		panicRef(c, err)
 	}
 
-	finalizerPtr := (*Finalizer)(atomic.LoadPointer(&c.finalizer))
-	if finalizerPtr != nil {
-		finalizer := *finalizerPtr
-		finalizer.Finalize()
+	if f := c.Finalizer(); f != nil {
+		f.Finalize()
 	}
 }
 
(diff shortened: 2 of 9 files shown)


checked/bytes.go
```diff
@@ -54,6 +54,7 @@ func NewBytes(value []byte, opts BytesOptions) Bytes {
 		value: value,
 	}
 	b.SetFinalizer(b)
+	b.TrackObject(b.value)
```
robskillington: Can we add a global like `trackLeak` in `debug.go` so we can avoid doing this all the time and enable only when we want to? i.e. similar to in `ref.go`:

```go
// ref.go
if traceback {
  tracebackEvent(...)
}
```

```go
// debug.go
var (
  traceLeak = false
)

func SetTraceLeak(value bool) {
  traceLeak = value
}
```

robskillington: Ah I see you do this with an atomic load.  Perhaps just a bool to avoid slightly slower atomic load?  We really should only enable/disable this at the very start of runtime.  That way we can avoid data races by only using the setters at runtime startup.

checked/ref.go
```diff
@@ -169,3 +172,73 @@ func (c *RefCount) DecWrites() {
...
+	})
+}
+
+// EnableLeakDetection turns leak detection on.
+func EnableLeakDetection() {
+	atomic.StoreUint64(&_LeakDetectionFlag, 1)
+}
+
+// DisableLeakDetection turns leak detection off.
+func DisableLeakDetection() {
```
robskillington: Could we make these methods similar to the other enable/disable methods in `debug.go`? i.e.

```go
func SetTraceLeak(value bool) {
  traceLeak = value
}
```

checked/ref.go
```diff
@@ -169,3 +172,73 @@ func (c *RefCount) DecWrites() {
...
+
+var _LeakDetectionFlag uint64
+
+var _Leaks struct {
+	sync.RWMutex
+	M map[string]uint64
+}
+
+func init() {
+	_Leaks.M = make(map[string]uint64)
```
robskillington: Any reason to use `_` named variables instead of just local vars like we do in `debug.go`?

checked/ref_test.go
```diff
@@ -204,3 +206,21 @@ func TestRefCountWriteFinishAfterFree(t *testing.T) {
...
+		v := &RefCount{}
+		v.TrackObject(v)
+		v.IncRef()
+	}
+
+	runtime.GC()
+
+	// Finalizers are run in a separate goroutine, so we have to wait
+	// a little bit here.
+	time.Sleep(100 * time.Millisecond)
```
robskillington: Can we put a while loop in here to make sure this test doesn't flap?

```go
for len(DumpLeaks()) == 0 {
  time.Sleep(100*time.Millisecond)
}
```

checked/bytes.go
kobolog: I can imagine one would want to trace leaks in a specific part of the code, by placing `EnableLeakDetection()` at the beginning of a section and `DisableLeakDetection()` at the end.

checked/ref.go
kobolog: I think Enable/Disable naming is superior in terms of readability – but I can rename functions in `debug.go` though.

checked/ref.go
kobolog: No reason, just experimenting with variable naming art.

checked/bytes.go
robskillington: That wouldn't work necessarily though as there could easily be concurrent goroutines that it would have side affects on (i.e. you wouldn't just be turning it on for the code between those calls, other code could receive the side affects too).

checked/ref.go
robskillington: Sure, fine with renaming it.  If you could move these enable/disable methods into `debug.go` too that would be more consistent.

checked/bytes.go
kobolog: Yes, it's a whole-program flag. If one knew though that there are leaks happening during a certain operation in a certain package, one could use these functions to enable instrumentation only when suspicious code is being executed.

checked/ref.go
```diff
@@ -169,3 +171,37 @@ func (c *RefCount) DecWrites() {
 func (c *RefCount) NumWriters() int {
 	return int(atomic.LoadInt32(&c.writes))
 }
+
+// TrackObject sets up the initial internal state of the Ref for
+// leak detection.
+func (c *RefCount) TrackObject(v interface{}) {
+	if atomic.LoadUint64(&leakDetectionFlag) == 0 {
+		return
+	}
```
robskillington: May it be better to do this before the `TrackObject(...)` call rather than inside it?  Performing the extra call stack and also creating a stack based `interface{}` every time even when we don't actually want to perform any tracking seems like an expense we could avoid pretty easily.

robskillington: Could export a method `ShouldTrackObject() bool` that performs this so others outside the module can use it?

kobolog: This proposed API seems backwards – the natural impulse for anyone reading such code would be to move the `ShouldTrackObject()` check _into_ the `TrackObject()` since there's no other use for that function, also it's a little bit too late to think about performance, is it not? Of course, I also hope that this whole debugging thing is a temporary intstrumentation so making a few unexpected code design choices scoped to it is fine by me if you insist.

kobolog: So do you insist or not?

PR #43, opened 2017-01-26
Add DebugMutex

kobolog:
WIP. This PR adds a `DebugMutex` type – a special version of `sync.RWMutex` that tracks its (exclusive, for now) ownership.

### DIFF
--- a/sync/debug_mutex.go
+++ b/sync/debug_mutex.go
@@ -8,29 +8,25 @@ import (
 	"time"
 )
 
-const _StackDepth = 16
-
 // DebugMutex is a RWMutex that tracks its ownership.
-type DebugMutex struct {
-	m sync.RWMutex
-}
+type DebugMutex sync.RWMutex
 
 // RLock locks DebugMutex for reading.
-func (m *DebugMutex) RLock() { m.m.RLock() }
+func (m *DebugMutex) RLock() { (*sync.RWMutex)(m).RLock() }
 
 // RUnlock undoes a single RLock call.
-func (m *DebugMutex) RUnlock() { m.m.RUnlock() }
+func (m *DebugMutex) RUnlock() { (*sync.RWMutex)(m).RUnlock() }
 
 // Lock locks DebugMutex for writing.
 func (m *DebugMutex) Lock() {
-	m.m.Lock()
+	(*sync.RWMutex)(m).Lock()
 	insert(m)
 }
 
 // Unlock unlocks DebugMutex for writing.
 func (m *DebugMutex) Unlock() {
 	remove(m)
-	m.m.Unlock()
+	(*sync.RWMutex)(m).Unlock()
 }
 
 // RLocker returns a Locker interface implemented via calls to RLock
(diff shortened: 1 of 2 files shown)


sync/debug_mutex.go
```diff
@@ -0,0 +1,123 @@
...
+// Unlock unlocks DebugMutex for writing.
+func (m *DebugMutex) Unlock() {
+	remove(m)
+	(*sync.RWMutex)(m).Unlock()
+}
+
+// RLocker returns a Locker interface implemented via calls to RLock
+// and RUnlock.
+func (m *DebugMutex) RLocker() sync.Locker {
+	return (*rlocker)(m)
```
xichen2020: Maybe just return `(*sync.RWMutex)(m).RLocker()` here so we don't need a separate `rlocker` type?

sync/debug_mutex.go
```diff
@@ -0,0 +1,123 @@
...
+func DisableMutexDebugging() {
+	mutexDebuggingFlag = false
+}
+
+// EnableMutexDebugging turns mutex debugging on.
+func EnableMutexDebugging() {
+	mutexDebuggingFlag = true
+}
+
+const _StackDepth = 16
```
xichen2020: Curious why `_` instead of lower casing `S`?

sync/debug_mutex.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	locks.Lock()
+	delete(locks.m, m)
+	locks.Unlock()
+}
+
+func traceback(l []uintptr) string {
+	var (
+		b    = new(bytes.Buffer)
```
xichen2020: nit: can just do `var b bytes.Buffer` to avoid allocating the struct

sync/debug_mutex.go
```diff
@@ -0,0 +1,123 @@
...
+import (
+	"bytes"
+	"fmt"
+	"runtime"
+	"sync"
+	"time"
+)
+
+// DebugMutex is a RWMutex that tracks its ownership.
+type DebugMutex sync.RWMutex
```
robskillington: Perhaps `DebugRWMutex`?

robskillington:
@kobolog could we potentially also tackle the problem of a rogue read lock being held blocking new writers?

kobolog:
Hm, apparently there is a tracing mutex already in debug, why am I doing this? O_o

sync/debug_mutex.go
kobolog: Oh, well, I can just to &b, right.

sync/debug_mutex.go
kobolog: Just want a way to differentiate between consts and global vars.

sync/debug_mutex.go
kobolog: Then it won't call my tracking `RLock()`/`RUnlock()`.

robskillington:
Yeah let's delete the one in debug, move this one into checked and set the ability to turn on/off the tracebacks.  Would be good as we briefly discussed, to detect certain bad situations like locks piling up over time, without the expense of tracebacks for use in production.

PR #45, opened 2017-02-07
Add use before initialization checks to the ObjectPool

robskillington:
cc @xichen2020 @cw9

### DIFF
--- a/pool/types.go
+++ b/pool/types.go
@@ -56,6 +56,10 @@ type CheckedObjectPool interface {
 	Get() checked.ReadWriteRef
 }
 
+// OnPoolAccessError is a function to call when a pool access error occurs,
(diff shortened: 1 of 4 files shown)


pool/types.go
```diff
@@ -56,6 +56,10 @@ type CheckedObjectPool interface {
 	Get() checked.ReadWriteRef
 }
 
+// OnPoolAccessError is a function to call when a pool access error occurs,
+// such as get or put before the pool is initialized.
+type OnPoolAccessError func(err error)
```
xichen2020: I think the convention is to name is `OnPoolAccessErrorFn`

robskillington: Sure thing.

PR #49, opened 2017-03-14
Add unsafe package to perform zero-allocation conversion from string to byte slice

xichen2020:
cc @kobolog @robskillington @jeromefroe @cw9 @prateek 

This PR adds the unsafe package to perform zero-allocation conversion from string to byte slices, which will significantly reduce GC when we need to encode string ids as byte slices.

### DIFF
--- a/unsafe/string.go
+++ b/unsafe/string.go
@@ -1,0 +1,50 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package xunsafe
+
+import (
+	"reflect"
+	"unsafe"
+)
+
+// ImmutableBytes represents an immutable byte slice
+type ImmutableBytes []byte
+
+// ToBytes converts a string to a byte slice with zero heap memory allocations.
+// The returned byte slice shares the same memory space as the string passed in.
+// It is the caller's responsibility to make sure the returned byte slice is not
+// modified in any way until end of life.
+func ToBytes(s string) ImmutableBytes {
+	if len(s) == 0 {
+		return nil
+	}
+	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
+
+	// NB(xichen): it is important that we access s in the same expression
+	// where we assign the data pointer of the string header to the data
+	// pointer of the slice header to make sure the underlying bytes don't
+	// get GC'ed before the assignment happens.
+	return *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
+		Data: sh.Data,
+		Len:  len(s),
+		Cap:  len(s),
+	}))
+}
--- a/unsafe/string.go
+++ b/unsafe/string.go
@@ -25,7 +25,7 @@ import (
 	"unsafe"
 )
 
-// ImmutableBytes represents an immutable byte slice
+// ImmutableBytes represents an immutable byte slice.
 type ImmutableBytes []byte
 
 // ToBytes converts a string to a byte slice with zero heap memory allocations.
@@ -36,15 +36,26 @@ func ToBytes(s string) ImmutableBytes {
 	if len(s) == 0 {
 		return nil
 	}
-	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
 
-	// NB(xichen): it is important that we access s in the same expression
-	// where we assign the data pointer of the string header to the data
-	// pointer of the slice header to make sure the underlying bytes don't
-	// get GC'ed before the assignment happens.
-	return *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
-		Data: sh.Data,
-		Len:  len(s),
-		Cap:  len(s),
-	}))
+	// NB(xichen): we need to declare a real byte slice so internally the compiler
+	// knows to use an unsafe.Pointer to keep track of the underlying memory so tha
+	// once the slice's array pointer is updated with the pointer to the string's
+	// underlying bytes, the compiler won't prematurely GC the memory when the string
+	// goes out of scope.
+	var b []byte
+	byteHeader := (*reflect.SliceHeader)(unsafe.Pointer(&b))
+
+	// NB(xichen): this makes sure that even if GC relocates the string's underlying
+	// memory after this assignment, the corresponding unsafe.Pointer in the internal
+	// slice struct will be updated accordingly to reflect the memory relocation.
+	byteHeader.Data = (*reflect.StringHeader)(unsafe.Pointer(&s)).Data
+
+	// NB(xichen): it is important that we access s after we assign the Data
+	// pointer of the string header to the Data pointer of the slice header to
+	// make sure the string (and the underlying bytes backing the string) don't get
+	// GC'ed before the assignment happens.
+	l := len(s)
+	byteHeader.Len = l
+	byteHeader.Cap = l
+	return b
 }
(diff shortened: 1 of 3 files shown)


unsafe/string.go
```diff
@@ -0,0 +1,50 @@
...
+
+// ToBytes converts a string to a byte slice with zero heap memory allocations.
+// The returned byte slice shares the same memory space as the string passed in.
+// It is the caller's responsibility to make sure the returned byte slice is not
+// modified in any way until end of life.
+func ToBytes(s string) ImmutableBytes {
+	if len(s) == 0 {
+		return nil
+	}
+	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
```
kobolog: You actually have to do all casts (`string` -> `StringHeader` -> `SliceHeader` -> `[]byte`) in the same expression, because currently after you cast `s` to `StringHeader`, a potential GC from the future version of Go could preempt the goroutine and move the pointer in `s`, while `sh`'s uintptr will be left unchanged.

jeromefroe: This may not actually come to fruition. There's a [proposal](https://github.com/golang/go/issues/19367) to add two new types `unsafe.String` and `unsafe.Slice` to replace `reflect.StringHeader` and `reflect.SliceHeader`. The `Data` field of the new types would be `unsafe.Pointer` instead of `uintptr` and `unsafe.Pointer` types will be updated by GC from what I can gather on this [thread](https://groups.google.com/forum/#!msg/golang-dev/rd8XgvAmtAA/p6r28fbF1QwJ) . The proposal is marked Go2 though so I gathers it's not very high on the list of priorities. With that being said, I also don't know if Go will have a compacting GC anytime soon.

unsafe/string.go
```diff
@@ -0,0 +1,50 @@
...
+
+	// NB(xichen): it is important that we access s in the same expression
+	// where we assign the data pointer of the string header to the data
+	// pointer of the slice header to make sure the underlying bytes don't
+	// get GC'ed before the assignment happens.
+	return *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
```
jeromefroe: The unsafe package states: "In general, reflect.SliceHeader and reflect.StringHeader should be used only as *reflect.SliceHeader and *reflect.StringHeader pointing at actual slices or strings, never as plain structs. A program should not declare or allocate variables of these struct types." And they offer the following pattern as being a valid way to update a string's `Data` field:

```
var s string
hdr := (*reflect.StringHeader)(unsafe.Pointer(&s)) // case 1
hdr.Data = uintptr(unsafe.Pointer(p))              // case 6 (this case)
hdr.Len = n
```

Consequently, I think we may be better off doing something of the form:

```
var b []byte
byteHeader := (*reflect.SliceHeader)(unsafe.Pointer(&b))
byteHeader.Data = (*reflect.StringHeader)(unsafe.Pointer(&s)).Data

l := len(s)
byteHeader.Len = l
byteHeader.Cap = l
return b
```

This way we:

1. Address Andrey's comment above by setting the `Data` field of the byte slice to be the same  as the `Data` field of the string in one expression.
2. Ensure `s` is live after updating the `Data` fields by calling `len(s)`.

kobolog: My comment still stays here, since there's a gap between reading the pointer from string header and writing a pointer to slice header, which potentially in the future can be used by the GC to preempt the goroutine. You literally need to read and write in the same expression, e.g.:

```go
return *(*[]byte)(unsafe.Pointer(&reflect.SliceHeader{
    Data: (*reflect.StringHeader)(unsafe.Pointer(&s)).Data,
    Cap: len(s),
    Len: len(s),
}))
```

kobolog: Also I'd like to point out that since we dynamically create a slice here, its memory won't be tracked by the GC and if the "underlying" string gets collected at any point in time while the slice is still being used, then the slice will end up having a dangling reference leading to all kinds of weird behavior.

jeromefroe: I fail to see how you're comment applies here, why would `Data: (*reflect.StringHeader)(unsafe.Pointer(&s)).Data,` be okay but not `byteHeader.Data = (*reflect.StringHeader)(unsafe.Pointer(&s)).Data`? It appears equivalent to your example to me since afterwards we have a byte slice pointing to the memory.

I also don't see why the memory wouldn't be tracked by GC. When the GC looks at the root pointers on the stack it will see a byte slice pointing to the memory that used to be pointed to by the string.

jeromefroe: What I mean to say is that if the GC can interrupt the load and subsequent store in the expression `byteHeader.Data = (*reflect.StringHeader)(unsafe.Pointer(&s)).Data` then surely it could do the same when we create the struct and do `Data: (*reflect.StringHeader)(unsafe.Pointer(&s)).Data,`.

kobolog: `byteHeader.Data` is a `uintptr` so it won't be updated if GC decides to relocate string's memory. In my example, we also immediately cast SliceHeader to a []byte that uses an `unsafe.Pointer` instead (https://golang.org/src/runtime/slice.go#L11) which is visible to the GC. Memory won't be tracked because type information for the GC is generated by the compiler, and this line of code will generate type information for `SliceHeader` because this is what we allocate. `SliceHeader` doesn't have any pointers so the heap area where it ends up will be marked as non-trackable. We **cast** that memory area to a `[]byte` later, but it doesn't matter.

jeromefroe: > byteHeader.Data is a uintptr so it won't be updated if GC decides to relocate string's memory.

Agreed, but we don't cast the `byteHeader` into a byte slice. We use it to update `b` which is a valid byte slice. Also, this pattern is explicitly marked as a valid pattern in the unsafe package, whereas creating a string or slice through their respective reflect Header structs is explicitly called out as something that one should not do.

> In my example, we also immediately cast SliceHeader to a []byte that uses an unsafe.Pointer instead (https://golang.org/src/runtime/slice.go#L11) which is visible to the GC.

The same is true here since the `unsafe.Pointer` in `b` will be updated by the line `byteHeader.Data = (*reflect.StringHeader)(unsafe.Pointer(&s)).Data`. We are in effect using `byteHeader` to access the private `array` field of a `slice` struct.

>  Memory won't be tracked because type information for the GC is generated by the compiler, and this line of code will generate type information for SliceHeader because this is what we allocate. SliceHeader doesn't have any pointers so the heap area where it ends up will be marked as non-trackable. We cast that memory area to a []byte later, but it doesn't matter.

Agreed, the memory will not be tracked by the `SliceHeader` but it will be tracked by the string `s` which is still live after the call into the `SliceHeader`'s `Data` field.

xichen2020: hm so it sounds like with all this trickery the byte slice might still get GC'ed if the string goes out of scope because we only allocated a slice header? 

If we are worried about it, would the following be better given that we are casting the underlying memory to a pointer to an array and take a slice based on that? Will the compiler see that we are taking a slice here and therefore won't release the memory backing the slice prematurely?
```
(*[0x7fffffff]byte)(unsafe.Pointer((*reflect.StringHeader)(unsafe.Pointer(&s)).Data))[:len(s):len(s)]
```

jeromefroe: I don't see why it would get GC'ed since the string is still live after we've updated the byte slice to point to the string's memory. After the string is last referenced the byte slice has already been updated to point to the memory so there isn't any point where there's no valid pointer to the memory. 

Also, I did some more digging and the [xxhash implementation that Influxdb uses](https://github.com/cespare/xxhash) takes the former approach instead of the latter:

```
func Sum64String(s string) uint64 {
	var b []byte
	sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
	bh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
	bh.Data = uintptr(unsafe.Pointer(sh.Data))
	bh.Len = sh.Len
	bh.Cap = sh.Len
	return Sum64(b)
}
```

kobolog: Yeah, you're correct that if you allocate a real slice and explicitly assign its `Data` member to whatever the string has then the GC will see two pointers to the same memory and if the string goes away, the memory will still be available. But in this xxhash example you provided, this hypothetical situation in the future Go version can happen:

```go
var b []byte
sh := (*reflect.StringHeader)(unsafe.Pointer(&s))
bh := (*reflect.SliceHeader)(unsafe.Pointer(&b))
// GC preempts this goroutine & relocates string's memory; sh.Data is a uintptr so it doesn't get updated and is a dangling pointer now.
bh.Data = uintptr(unsafe.Pointer(sh.Data))
bh.Len = sh.Len
bh.Cap = sh.Cap
return Sum64(b) // Reading invalid memory here.
```

kobolog: To clarify, my comment about dangling pointers applies to all examples and is valid even in the light of the documentation code snippet provided by Jerome since `p` there (on line 3) is a **valid pointer all along**, unlike `sh.Data` in our examples.
My comment about GC tracking memory after the string gets collected applies to the original code, not the proposed changes where we start with allocating a real `[]byte`.

jeromefroe: I agree that there is a problem in the xxhash implementation, I offered it just to highlight the fact that it takes the approach of allocating a real byte slice (i.e. `var b []byte`) and not allocating a byte slice through the use of a `SliceHeader`.

jeromefroe: I also agree about the possibility of a dangling pointer, but it will only be an issue for a GC that moves memory and updates references. There is already a proposal to introduce a new unsafe representation for slices and strings whose `Data` field would be an `unsafe.Pointer` instead of an `uintptr`. If that proposal lands before a compacting GC is introduced then we can update this package to use those new types and won't run into this issue even with a compacting GC. Needless to say, we'll need to be vigilant during new Go releases.

jeromefroe: Hmm, actually, on second thought, we should be fine even if the GC begins moving references with this approach because what we are doing is using the reflect Header structs to update the internal `unsafe.Pointer`s in the runtime representation of strings and slices.

xichen2020: I don't see why we'd have a dangling pointer. If GC relocates the string's underlying memory, it'll need to fix up the str field in the stringStruct object on the stack:
```
type stringStruct struct {
  str unsafe.Pointer
  len int
}
```
If that's the case, it means it'll automatically fix up the Data pointer in the string header because the string header pointer is literally just reinterpreting cast the stringStruct pointer and points to the exact same location on the stack where the stringStruct variable sits.

PR #50, opened 2017-03-15
Add implementation of jump consistent hash algorithim

jeromefroe:
@robskillington @ben-lerner

### DIFF
--- a/hash/jump/jump_test.go
+++ b/hash/jump/jump_test.go
@@ -34,18 +34,18 @@ func TestHashDistribution(t *testing.T) {
 
 	numBuckets := []int64{10, 100, 1000, 10000}
 	for i, n := range numBuckets {
-		keysPerBucket := uint64(n) * 1000
-		stepSize := math.MaxUint64 / keysPerBucket
+		totalKeys := uint64(n) * 1000
+		stepSize := math.MaxUint64 / totalKeys
 		buckets := make(map[int64]int, n)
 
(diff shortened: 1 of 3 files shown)


hash/jump/jump_test.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+}
+
+// TestHashMoved tests that Hash only redistributes approximately 1/(n+1) keys when going from
+// n to n+1 buckets
+func TestHashMoved(t *testing.T) {
+	numBuckets := []int64{10, 100, 1000}
+
+	for _, n := range numBuckets {
+		maxNum := uint64(n) * 1000
```
robskillington: Instead of choosing `[0, n*1000)` to test I would instead check `[0, math.MaxUint64)` with a step size of `math.MaxUint64 / (n * 1000)`.  This will test the entire keyspace instead of a subset.

jeromefroe: I picked `n*1000` just to ensure we have a large number of keys per bucket. I also like the idea of spreading the keys out more though. What if we change the bucket sizes to be `{2^4, 2^8, 2^12}` and set the step sizes to `math.MaxUint64 >> (10+ln(n))`? This will give us about `2^10` keys per bucket again while also ensuring they span the whole range of keys.

jeromefroe: Disregard my previous comment, I misread your comment.

PR #51, opened 2017-03-24
Add unsafe function to convert a byte slice to string with no heap allocation

xichen2020:
cc @jeromefroe @kobolog @robskillington 

This PR adds an unsafe method to convert a byte slice to a string with zero heap allocation.

### DIFF
--- a/unsafe/bytes.go
+++ b/unsafe/bytes.go
@@ -1,0 +1,60 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


kobolog:
Apparently, the lifetime test is failing. It happens either because of some bug in codegen or because we somehow fundamentally misunderstood the way GC & heap bitmaps work.

xichen2020:
Turns out sometimes the byte slice could be allocated on the stack per compiler optimization, in which case if the byte slice goes out of scope, the string will point to somewhere on the stack and may be overwritten by function calls later. In general, there is no good way of checking or making sure the byte slice is allocated on the heap, which mean this function is only valid for as long as the string lives. This could turn out to be confusing to the users of this function, so for now I'll avoid adding it and may revisit the need in the future.

PR #54, opened 2017-04-11
fix panic in NullLogger

cw9:
There was a panic when calling NullLogger.WithFields() because the
fields was nill

@xichen2020 @robskillington @prateek

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -142,13 +142,13 @@ func (nullLogger) Debug(msg string)                       {}
 func (l nullLogger) Fields() LogFields                    { return l.fields }
 
 func (l nullLogger) WithFields(newFields ...LogField) Logger {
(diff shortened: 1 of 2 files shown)


log/logger.go
```diff
@@ -122,7 +122,7 @@ func (f Fields) Len() int { return len(f) }
 func (f Fields) ValueAt(i int) LogField { return f[i] }
 
 // NullLogger is a logger that emits nowhere
-var NullLogger Logger = nullLogger{}
+var NullLogger Logger = nullLogger{fields: Fields([]LogField{})}
```
xichen2020: Perhaps we should just check if the fields is nil and avoid calling `Len` if it is to avoid creating a slice?

cw9: yeah good call

PR #56, opened 2017-04-13
Add config parser

cw9:
@xichen2020 @prateek @robskillington

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -1,0 +1,35 @@
+package config
+
+import (
+	"errors"
+	"io/ioutil"
+
+	validator "gopkg.in/validator.v2"
+	yaml "gopkg.in/yaml.v2"
+)
+
+// errNoFilesToLoad is return when you attemp to call LoadFiles with no file paths
+var errNoFilesToLoad = errors.New("attempt to load configuration with no files")
+
+// LoadFiles loads a list of files, deep-merging values.
+// Validation is done after merging all values
+func LoadFiles(config interface{}, fnames ...string) error {
+	if len(fnames) == 0 {
+		return errNoFilesToLoad
+	}
+	for _, fname := range fnames {
+		data, err := ioutil.ReadFile(fname)
+		if err != nil {
+			return err
+		}
+
+		if err := yaml.Unmarshal(data, config); err != nil {
+			return err
+		}
+	}
+
+	if err := validator.Validate(config); err != nil {
+		return err
+	}
+	return nil
+}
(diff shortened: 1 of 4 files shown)


config/config.go
```diff
@@ -0,0 +1,35 @@
...
+		if err != nil {
+			return err
+		}
+
+		if err := yaml.Unmarshal(data, config); err != nil {
+			return err
+		}
+	}
+
+	if err := validator.Validate(config); err != nil {
```
xichen2020: nit: `return validator.Validate(config)`

config/config.go
```diff
@@ -0,0 +1,35 @@
...
+	if len(fnames) == 0 {
+		return errNoFilesToLoad
+	}
+	for _, fname := range fnames {
+		data, err := ioutil.ReadFile(fname)
+		if err != nil {
+			return err
+		}
+
+		if err := yaml.Unmarshal(data, config); err != nil {
```
xichen2020: What's the expected behavior if multiple files provide different values for the same config?

cw9: If value for a property is present in multiple files, the value from the last file will be applied

Updated the comment as well

config/config.go
cw9: +1

config/config.go
prateek: Do we have a use-case in the codebase which relies on merging yamls? If not, I'd say we don't expose that functionality.

cw9: We don't right now, my thinking is that with this ability we can have a default yaml file in OSS repo and have another override yaml in the statsdex repo that only overrides a few fields.

xichen2020: I'm against merging unless it's absolutely necessary because it makes it difficult to reason about what the final values are for each config property without reading every config. It also means we now have one more config file we need to manage. AFAICT we've always been using a single config, so we can probably live without merging.

That said, I'm not against making the `LoadFiles` function unexported and have a `LoadFile` function that interally calls `LoadFiles` passing in a single file. That way if one day we do need the merging ability, we can easily export it.
  
Also, what are your thoughts on defining a simple mechanism to find default config files (e.g., always look inside the current working dir for `config.yaml`) and providing a function (e.g., `Load`) to load the default config file if it exists?

cw9: Sounds good, I'll only expose LoadFile() for now and hide the LoadFiles(). For the default config files, it does not seem super necessary given that we always run the main with a config file passed in, gonna punt it for now and can add if needed

PR #59, opened 2017-04-16
Add instrumentation and retry configurations

xichen2020:
cc @jeromefroe @martin-mao @cw9  @robskillington 

This PR adds metrics configurations to `instrument/` for reuse across repos.

### DIFF
--- a/instrument/config.go
+++ b/instrument/config.go
@@ -1,4 +1,4 @@
-// Copyright (c) 2016 Uber Technologies, Inc.
+// Copyright (c) 2017 Uber Technologies, Inc.
 //
 // Permission is hereby granted, free of charge, to any person obtaining a copy
 // of this software and associated documentation files (the "Software"), to deal
--- a/instrument/config.go
+++ b/instrument/config.go
@@ -29,27 +29,32 @@ import (
 	"github.com/uber-go/tally/m3"
 )
 
-const (
-	defaultSamplingRate      = 1.0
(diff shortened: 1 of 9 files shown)


instrument/config.go
```diff
@@ -0,0 +1,116 @@
+// Copyright (c) 2016 Uber Technologies, Inc.
```
cw9: nit: 2017?

instrument/config.go
```diff
@@ -0,0 +1,116 @@
...
+// SampleRate returns the metrics sampling rate.
+func (mc *MetricsConfiguration) SampleRate() float64 {
+	if mc.SamplingRate > 0.0 && mc.SamplingRate <= 1.0 {
+		return mc.SamplingRate
+	}
+	return defaultSamplingRate
+}
+
+// ReportInterval returns the metrics reporting interval.
+func (mc *MetricsConfiguration) ReportInterval() time.Duration {
```
cw9: does this need to be exported?

instrument/config.go
xichen2020: Yeah good catch, will fix.

instrument/config.go
xichen2020: Yeah the instrumentation options have a `ReportInterval` property that's used to control how frequently gauge metrics are reported in separate goroutines (e.g., number of file descriptors, and all the database related gauge metrics in m3db).

PR #60, opened 2017-04-17
Add pprof handler for custom http mux

xichen2020:
cc @jeromefroe @cw9 @prateek @martin-mao 

This PR adds the pprof handler for custom http mux.

### DIFF
--- a/pprof/pprof.go
+++ b/pprof/pprof.go
@@ -1,0 +1,55 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


pprof/pprof.go
```diff
@@ -0,0 +1,55 @@
...
+	"strings"
+)
+
+const (
+	pprofPath = "/debug/pprof/"
+)
+
+// RegisterHandler registers the pprof handler with the given http mux.
+func RegisterHandler(mux *http.ServeMux) {
+	mux.Handle(pprofPath, handler())
```
prateek: more a nit: is there a case the user would want to configure the endpoint url too?

xichen2020: Don't think that's a common usecase, we tend to (and the standard library recommends) using `/debug/pprof`.

prateek: cool, fair enough

PR #62, opened 2017-04-29
Report extended metrics

xichen2020:
cc @robskillington @cw9 @jeromefroe @prateek 

This PR adds ability to report runtime metrics.

### DIFF
--- a/instrument/config.go
+++ b/instrument/config.go
@@ -56,8 +56,8 @@ type MetricsConfiguration struct {
 	// Metrics sampling rate.
 	SamplingRate float64 `yaml:"samplingRate" validate:"nonzero,min=0.0,max=1.0"`
 
-	// Whether to disable reporting runtime metrics.
-	DisableRuntimeMetrics bool `yaml:"disableRuntimeMetrics"`
+	// Whether to disable reporting extended metrics.
+	DisableExtendedMetrics bool `yaml:"disableExtendedMetrics"`
 }
 
 // NewRootScope creates a new tally.Scope based on a tally.CachedStatsReporter
@@ -70,8 +70,7 @@ func (mc *MetricsConfiguration) NewRootScope() (tally.Scope, io.Closer, error) {
 	if err != nil {
 		return nil, nil, err
 	}
-	s, c := mc.NewRootScopeReporter(r)
-	return s, c, nil
+	return mc.NewRootScopeReporter(r)
 }
 
 // NewRootScopeReporter creates a new tally.Scope based on a given tally.CachedStatsReporter
@@ -79,7 +78,7 @@ func (mc *MetricsConfiguration) NewRootScope() (tally.Scope, io.Closer, error) {
 // as hooking into the reporter to manually flush it.
 func (mc *MetricsConfiguration) NewRootScopeReporter(
 	r tally.CachedStatsReporter,
-) (tally.Scope, io.Closer) {
+) (tally.Scope, io.Closer, error) {
 	var (
 		prefix string
 		tags   map[string]string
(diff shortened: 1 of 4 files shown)


instrument/runtime.go
```diff
@@ -0,0 +1,145 @@
...
+			NumGC:           memoryScope.Counter("num-gc"),
+			GcPauseMs:       memoryScope.Timer("gc-pause-ms"),
+			lastNumGC:       memstats.NumGC,
+		},
+		started: false,
+		quit:    make(chan struct{}),
+	}
+}
+
+// Start Starts the reporter thread that periodically emits metrics.
```
jeromefroe: nit: "Starts" shouldn't be capitalized.

instrument/config.go
```diff
@@ -96,7 +99,15 @@ func (mc *MetricsConfiguration) NewRootScopeReporter(
 		Prefix:         prefix,
 		CachedReporter: r,
 	}
-	return tally.NewRootScope(scopeOpts, mc.ReportInterval())
+	reportInterval := mc.ReportInterval()
+	scope, closer := tally.NewRootScope(scopeOpts, reportInterval)
+
+	if !mc.DisableRuntimeMetrics {
+		runtimeScope := scope.SubScope("runtime")
```
jeromefroe: I'm wondering if we want to use a `Tagged` subscope here, e.g. `scope.Tagged(map[string]string{"source": "runtime"})`, to avoid concatenating the names of the subscopes. We've been encouraging other teams to take this approach and it might offer us additional flexibility.

instrument/runtime.go
```diff
@@ -0,0 +1,145 @@
...
+
+// NewRuntimeMetricsReporter Creates a new RuntimeMetricsReporter.
+func NewRuntimeMetricsReporter(
+	scope tally.Scope,
+	reportInterval time.Duration,
+) *RuntimeMetricsReporter {
+	var memstats runtime.MemStats
+	runtime.ReadMemStats(&memstats)
+
+	memoryScope := scope.SubScope("memory")
```
jeromefroe: Same comment as above, might be worth considering using a `Tagged` subscope here, something like `scope.Tagged(map[string]string{"category": "memory"})`.

instrument/runtime.go
```diff
@@ -0,0 +1,145 @@
...
+			case <-r.quit:
+				ticker.Stop()
+				return
+			}
+		}
+	}()
+	r.started = true
+}
+
+// Stop Stops reporting of runtime metrics.
```
jeromefroe: nit: "Stops" should be "stops".

instrument/runtime.go
```diff
@@ -0,0 +1,145 @@
...
+// The reporter cannot be started again after it's been stopped.
+func (r *RuntimeMetricsReporter) Stop() {
+	close(r.quit)
+}
+
+// report Sends runtime metrics to the local metrics collector.
```
jeromefroe: nit: same as above, might want to lowercase "Sends".

instrument/runtime.go
```diff
@@ -0,0 +1,182 @@
...
+			process:     process,
+		},
+		started: false,
+		quit:    make(chan struct{}),
+	}, nil
+}
+
+// Start starts the reporter thread that periodically emits metrics.
+func (r *ExtendedMetricsReporter) Start() {
+	if r.started {
```
jeromefroe: nit: We may want to consider using a mutex here to protect `r.started` since there's still a possibility, albeit very small, of `Start` being called concurrently and the two report goroutines being kicked off.

jeromefroe:
LGTM other than small nit

instrument/runtime.go
xichen2020: Yeah the reporter is not meant to be used in a multithreaded environment. I'll add a comment.

PR #69, opened 2017-05-07
Add ability to specify extended metrics types

### LINKED ISSUE #64: Add ability to report simple runtime metrics to avoid STW pauses
cc @robskillington 

We currently report runtime metrics including memory/gc -related runtime metrics from instrument/config.go and instrument/extended.go. However because `runtime.ReadMemStats` requires a stop-the-world pause when gathering the heap stats, reporting memory/gc-related runtime metrics can cause perf degradations especially when they are reported frequently in processes with large heaps.

We should add an option to allow users to opt out of reporting those stats if those are deemed too expensive to have.

xichen2020:
cc @robskillington @cw9 @jeromefroe @prateek 

This PR adds ability for users to choose the level of details for extended metrics, in particular if reporting detailed memory-related metrics is deemed too expensive due to STW pauses and they'd like to stick with reporting simple extended metrics.

Fixes #64 .


instrument/extended.go
```diff
@@ -21,26 +21,99 @@
...
-	"github.com/shirou/gopsutil/process"
+	"github.com/m3db/m3x/process"
+
 	"github.com/uber-go/tally"
 )
 
+// ExtendedMetricsType is a type of extended metrics to report.
+type ExtendedMetricsType int
+
+const (
```
jeromefroe: Perhaps we might want to use a bitflag to determine which metrics to report or not so we have more fine-grained control in the future? I can see scenarios where we're testing builds of the collector where we want to report a certain metric that is contained in `DetailedExtendedMetrics` (for example, GC pause times) but don't need the other metrics. In this case, we can still have constants for no extended metrics,  simple extended metrics, and detailed extended metrics which will be equal to their respective bitflag integers, and be used in the majority of cases to simplify things without sacrificing the option to turn individual metrics on or off.

xichen2020: Talked offline, the simple metrics (goroutine count etc) are generally nice to have and cheap to report so leaving it as is for now.

process/process_fallback.go
```diff
@@ -0,0 +1,35 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// +build !linux
```
jeromefroe: nit: maybe just call this file `process.go`?

process/process_linux.go
```diff
@@ -0,0 +1,42 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// +build linux
```
jeromefroe: This is actually redundant, because the file ends with `_linux` the go compiler will only compile this file on Linux.

instrument/extended.go
```diff
@@ -21,26 +21,99 @@
...
+	case SimpleExtendedMetrics:
+		return "simple"
+	case DetailedExtendedMetrics:
+		return "detailed"
+	}
+	return "unknown"
+}
+
+// UnmarshalYAML unmarshals an ExtendedMetricsType into a valid type from string.
+func (t *ExtendedMetricsType) UnmarshalYAML(unmarshal func(interface{}) error) error {
```
jeromefroe: nit: maybe just add a comment specifying that we are implementing the `Unmarshaler` interface for the [yaml.v2](https://godoc.org/gopkg.in/yaml.v2) package.

jeromefroe:
LGTM besides the couple of nits.

process/process_fallback.go
xichen2020: sg, will do

process/process_linux.go
xichen2020: yeah but it's more robust to have it in here just in case this file is renamed to sth else in the future.

xichen2020: Removed the build flag to be consistent w/ the stdlib usage pattern.

PR #71, opened 2017-05-20
Move server from m3aggregator to m3x

cw9:
So it can be reused by aggregator, ingester etc.

@xichen2020

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -118,6 +118,7 @@ func (s *server) ListenAndServe() error {
 func (s *server) serve() error {
 	connCh, errCh := xnet.StartAcceptLoop(s.listener, s.opts.Retrier())
 	for conn := range connCh {
+		conn := conn
 		if !s.addConnectionFn(conn) {
 			conn.Close()
 		} else {
--- a/server/server.go
+++ b/server/server.go
@@ -92,7 +92,7 @@ func NewServer(address string, handler Handler, opts Options) Server {
 		address:    address,
 		opts:       opts,
 		log:        instrumentOpts.Logger(),
-		closedChan: make(chan struct{}, 1),
+		closedChan: make(chan struct{}),
 		metrics:    newServerMetrics(scope),
 		handler:    handler,
 	}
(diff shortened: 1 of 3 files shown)


server/server.go
```diff
@@ -0,0 +1,202 @@
...
+	handler  Handler
+
+	addConnectionFn    addConnectionFn
+	removeConnectionFn removeConnectionFn
+}
+
+// NewServer creates a new server.
+func NewServer(address string, handler Handler, opts Options) Server {
+	instrumentOpts := opts.InstrumentOptions()
+	scope := instrumentOpts.MetricsScope().SubScope("server")
```
xichen2020: The convention is usually to assume the scope passed in is the correct scope for the current object.

server/server.go
```diff
@@ -0,0 +1,202 @@
...
+				s.wgConns.Done()
+			}()
+		}
+	}
+	return <-errCh
+}
+
+func (s *server) Close() {
+	s.Lock()
+	if atomic.LoadInt32(&s.closed) == 1 {
```
xichen2020: Can probably change this to a bool

server/server.go
```diff
@@ -86,13 +87,14 @@ type server struct {
...
 	s := &server{
-		address: address,
-		opts:    opts,
-		log:     instrumentOpts.Logger(),
-		metrics: newServerMetrics(scope),
-		handler: handler,
+		address:    address,
+		opts:       opts,
+		log:        instrumentOpts.Logger(),
+		closedChan: make(chan struct{}, 1),
```
xichen2020: Could just be `make(chan struct{})`

server/server.go
```diff
@@ -193,10 +196,11 @@ func (s *server) reportMetrics() {
 	t := time.Tick(interval)
 
 	for {
-		<-t
-		if atomic.LoadInt32(&s.closed) == 1 {
+		select {
+		case <-t:
+			s.metrics.openConnections.Update(float64(atomic.LoadInt32(&s.numConns)))
+		case <-s.closedChan:
```
xichen2020: Could create a ticker at line 196 and stop the ticker here.

PR #72, opened 2017-05-23
Add Close function to server handler interface

cw9:
@xichen2020


server/server.go
```diff
@@ -49,6 +49,9 @@ type Handler interface {
 	// Handle handles the data received on the connection, this function
 	// should be blocking until the connection is closed or received error.
 	Handle(conn net.Conn)
+
+	// Close closes the handler
```
xichen2020: Add a period at the end.

PR #73, opened 2017-05-24
Add StartForeverLoop to support accepting connections in an infinite loop

xichen2020:
@robskillington @cw9

This PR adds `StartForeverLoop` to support accepting connections in an infinite loop.

### DIFF
--- a/retry/types.go
+++ b/retry/types.go
@@ -46,6 +46,10 @@ type ContinueFn func(attempt int) bool
 
 // Retrier is a executor that can retry attempts on executing methods
 type Retrier interface {
+	// SetForeverRetry sets the retry mode that determines whether attempts should
+	// be retried forever.
+	SetForeverRetry(forever bool) Retrier
+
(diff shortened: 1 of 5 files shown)


retry/types.go
```diff
@@ -46,6 +46,10 @@ type ContinueFn func(attempt int) bool
 
 // Retrier is a executor that can retry attempts on executing methods
 type Retrier interface {
+	// SetForeverRetry sets the retry mode that determines whether attempts should
+	// be retried forever.
+	SetForeverRetry(forever bool) Retrier
```
cw9: why can't we use the SetForever in Retry Options?

xichen2020: Because the server ctor takes in a `Retrier`. I could potentially expose the `Options` method, call `SetForever` on the options, and create a new retrier in `StartForeverAcceptLoop`, although I'll need to store the options in the retrier for that to work.

PR #74, opened 2017-05-24
[servers] Add Serve method to server interface

jeromefroe:
@xichen2020 @cw9 

This PR adds a `Serve(l net.Listener)` method to the `Server` interface so that we can explicitly pass a listener to a `Server`. This will be helpful in testing when we don't know in advance what port we want to listen on.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -124,7 +124,7 @@ func (s *server) ListenAndServe() error {
 }
 
 func (s *server) Serve(l net.Listener) error {
-	s.address := l.Addr().String()
+	s.address = l.Addr().String()
 	s.listener = l
 	go s.serve()
 	return nil
--- a/server/server.go
+++ b/server/server.go
@@ -124,8 +124,7 @@ func (s *server) ListenAndServe() error {
 }
 
 func (s *server) Serve(l net.Listener) error {
-	addr := l.Addr()
-	s.address = addr.String()
+	s.address := l.Addr().String()
 	s.listener = l
 	go s.serve()
 	return nil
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -120,6 +123,13 @@ func (s *server) ListenAndServe() error {
 	return nil
 }
 
+func (s *server) Serve(l net.Listener) error {
+	s.address = l.Addr().String()
+	s.listener = l
+	go s.serve()
+	return nil
```
cw9: do we need to return error in the function signature?

server/server_test.go
```diff
@@ -101,6 +101,18 @@ func TestServerListenAndClose(t *testing.T) {
...
 }
 
+func TestServe(t *testing.T) {
+	s, _, _, _ := testServer(testListenAddress)
+
+	l, err := net.Listen("tcp", testListenAddress)
+	require.NoError(t, err)
+
+	err = s.Serve(l)
+	require.NoError(t, err)
```
cw9: also verify the address and listener in s?

server/server.go
jeromefroe: It's not necessary in this case, but I can imagine we might have other implementations of the `Server` interface that might return an error. This way we wouldn't have to change the function signature in the future. I don't feel too strongly either way though.

server/server_test.go
jeromefroe: Sure thing.

server/server.go
xichen2020: +1, the http server has the same signature so we should follow suit

PR #75, opened 2017-05-25
Add ListenAndServeForever to m3x server

cw9:
Add this API as per discussion, although I feel we should just change ListenAndServe() and Serve() to do the Forever thing under the hood rather than adding this new API, because the user of the server interface has no way of knowing if the server has gone out of retries.

@xichen2020 @robskillington @jeromefroe @prateek

### DIFF
--- a/net/server.go
+++ b/net/server.go
@@ -30,11 +30,12 @@ import (
 // StartAcceptLoop starts an accept loop for the given listener,
 // returning accepted connections via a channel while handling
 // temporary network errors. Fatal errors are returned via the
-// error channel with the listener closed on return
-func StartAcceptLoop(l net.Listener, retrier xretry.Retrier) (<-chan net.Conn, <-chan error) {
+// error channel with the listener closed on return.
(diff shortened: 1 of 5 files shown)


net/server.go
```diff
@@ -31,10 +31,11 @@ import (
 // returning accepted connections via a channel while handling
 // temporary network errors. Fatal errors are returned via the
 // error channel with the listener closed on return
-func StartAcceptLoop(l net.Listener, retrier xretry.Retrier) (<-chan net.Conn, <-chan error) {
```
robskillington: Can you add another constructor that simply takes `l net.Listener` that does the forever loop by default?  i.e. 
```go
func StartForeverAcceptLoop(l net.Listener) (<-chan net.Conn, <-chan error) {
  return StartAcceptLoop(l, xretry.NewOptions().SetForever(true))
}
``` 

That way users when looking at the package don't need to know they need to set a special flag, they can just use the constructor out of the box.

cw9: sure, sg

PR #76, opened 2017-05-30
Add NewOptions() to retry config

cw9:
@xichen2020 @jeromefroe

### DIFF
--- a/retry/config.go
+++ b/retry/config.go
@@ -42,40 +42,35 @@ type Configuration struct {
 
 	// Whether to retry forever until either the attempt succeeds,
 	// or the retry condition becomes false.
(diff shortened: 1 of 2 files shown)


retry/config.go
```diff
@@ -68,13 +68,17 @@ func (c Configuration) NewRetrier(scope tally.Scope) Retrier {
 	if c.MaxRetries != 0 {
 		maxRetries = c.MaxRetries
 	}
-	retryOpts := NewOptions().
+	return NewOptions().
```
xichen2020: btw I know this is not your change, but `c.Forever` and `c.Jitter` should probably be *bool in the config, otherwise you can't use the default and always need to pick one

PR #79, opened 2017-06-14
Optionally support keepalives for tcp connections

xichen2020:
cc @cw9 

This PR adds support to optionally set keepAlives for TCP connections.

### DIFF
--- a/server/server.go
+++ b/server/server.go
@@ -28,6 +28,7 @@ import (
 
 	"github.com/m3db/m3x/log"
 	"github.com/m3db/m3x/net"
+	"github.com/m3db/m3x/retry"
 
 	"github.com/uber-go/tally"
 )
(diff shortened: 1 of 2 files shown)


server/server.go
```diff
@@ -96,12 +99,14 @@ func NewServer(address string, handler Handler, opts Options) Server {
...
 
 	s := &server{
-		address:    address,
-		opts:       opts,
-		log:        instrumentOpts.Logger(),
-		closedChan: make(chan struct{}),
-		metrics:    newServerMetrics(scope),
-		handler:    handler,
+		address:                address,
+		opts:                   opts,
```
cw9: no longer used?

xichen2020: Still used for report interval, but I can store it separately as well and get rid of opts.

PR #80, opened 2017-06-16
[election] Add package wrapping etcd's conc. pkg

schallert:
Behavior documented in comments

### DIFF
--- a/election/client.go
+++ b/election/client.go
@@ -14,11 +14,6 @@ import (
 )
 
 var (
-	// ErrCampaignInProgress is returned when a client tries to start a second
-	// camapaign if they are either (1) already the leader or (2) not the leader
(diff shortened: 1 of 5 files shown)


election/client.go
```diff
@@ -0,0 +1,292 @@
...
+		}
+	}
+
+	c.mu.RLock()
+	session := c.session
+	c.mu.RUnlock()
+
+	// if we lose the session in the background and that session was the
+	// client's active one we want to allow the client to campaign again
+	go func(session *concurrency.Session) {
```
cw9: what's the difference between this function and the select in line 139?

schallert:
Going to start this in m3cluster instead. If we end up implementing all `concurrency.Election` functions down the road we can move it back.

PR #81, opened 2017-06-27
Enable setting keepAlive period for tcp connections

xichen2020:
cc @cw9 

This enables setting keepAlive period for TCP connections in the server.

### DIFF
--- a/server/options.go
+++ b/server/options.go
@@ -21,12 +21,18 @@
 package xserver
 
 import (
+	"time"
+
 	"github.com/m3db/m3x/instrument"
 	"github.com/m3db/m3x/retry"
 )
 
 const (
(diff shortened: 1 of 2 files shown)


server/options.go
```diff
@@ -21,12 +21,18 @@
...
+
 	"github.com/m3db/m3x/instrument"
 	"github.com/m3db/m3x/retry"
 )
 
 const (
+	// By default keepAlives are enabled for TCP connections.
 	defaultTCPConnectionKeepAlive = true
+
+	// By default the keep alive period is not set.
```
cw9: also indicate that the default keep alive period in tcpConn will be used when no override value is provided?

xichen2020: Sure

PR #84, opened 2017-08-07
Add Range.Intersect()

prateek:
- Utility method to squeeze a range within another

/cc @martin-mao @xichen2020

### DIFF
--- a/time/range.go
+++ b/time/range.go
@@ -61,10 +61,10 @@ func (r Range) Overlaps(other Range) bool {
 	return r.End.After(other.Start) && r.Start.Before(other.End)
 }
 
-// SqueezeWithin squeezes the receiver `Range` to be contained within the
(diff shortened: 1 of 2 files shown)


time/range.go
```diff
@@ -61,6 +61,23 @@ func (r Range) Overlaps(other Range) bool {
 	return r.End.After(other.Start) && r.Start.Before(other.End)
 }
 
+// SqueezeWithin squeezes the receiver `Range` to be contained within the
+// provided argument range iff there is an overlap between the two. It also
+// returns a bool indicating if it was able to squeeze the range successfully.
+func (r Range) SqueezeWithin(other Range) (Range, bool) {
```
xichen2020: `Intersect` is probably a better name?

prateek: Definitely is, will change.

prateek: done

PR #85, opened 2017-08-17
[gometalinter] Add gometalinter

jeromefroe:
Just like previous work in other repos under the `m3db` organization, this PR adds `gometalinter` to the `m3x` repo and fixes any issues it finds.

cc @xichen2020 @prateek


Makefile
```diff
@@ -1,48 +1,68 @@
...
+.PHONY: test
 test: test-internal
 	gocov convert $(coverfile) | gocov report
 
+.PHONY: testhtml
 testhtml: test-internal
 	gocov convert $(coverfile) | gocov-html > $(html_report) && open $(html_report)
 	@rm -f $(test_log) &> /dev/null
 
+.PHONY: test-internal
```
prateek: typo, should be `test-ci-unit`

debug/mutex_test.go
```diff
@@ -65,21 +65,20 @@ func TestRWMutexReportEvery(t *testing.T) {
 	}()
 
 	// Wait for writer to appear
-	for {
-		if atomic.LoadInt64(&mutex.writers) != 1 {
-			time.Sleep(time.Millisecond)
-		}
-		break
+	for atomic.LoadInt64(&mutex.writers) != 1 {
```
prateek: Confirming that you intended to change semantics ~ the original code sleeps at most once, and then breaks out the loop. The new code is going to sleep repeatedly till the condition is true.

jeromefroe: Will update this, the previous should have never had a `for` loop, at least it was useless.

checked/debug.go
```diff
@@ -44,7 +44,11 @@ var (
 )
 
 var tracebackCallersPool = sync.Pool{New: func() interface{} {
-	return make([]uintptr, tracebackMaxDepth)
+	// Pools should generally only return pointer types, since a pointer
+	// can be put into the return interface value without an allocation.
+	// Consequently, we return a pointer to a slice instead of just a slice.
+	pc := make([]uintptr, tracebackMaxDepth)
+	return &pc
```
xichen2020: I don't think this is the right thing to do based on the usage pattern, see below.

checked/debug.go
```diff
@@ -189,7 +193,7 @@ func (d *debugger) append(event debuggerEvent, ref int, pc []uintptr) {
 			// Shift all tracebacks back one if at end of traceback cycles
 			slice := d.entries[0]
 			for i, entry := range slice {
-				tracebackCallersPool.Put(entry.pc)
+				tracebackCallersPool.Put(&entry.pc)
```
xichen2020: If you put back the address, and you nil it out immediately below, the allocated slice is lost, and when you get it back at line 297, you'll get a nil back, which probably isn't what we want.

jeromefroe: Yea, good point, we need to change the field in `entry` to be a pointer. Will update.

PR #88, opened 2017-08-18
[instrument] Add option to specify sanitization options in instrument package

jeromefroe:
This PR adds the ability to specify sanitization options in the `instrument` package so callers can use a scope which sanitizes their metrics before emission.

cc @xichen2020 @prateek


instrument/config.go
```diff
@@ -58,6 +58,9 @@ type MetricsConfiguration struct {
 
 	// Extended metrics type.
 	ExtendedMetrics *ExtendedMetricsType `yaml:"extended"`
+
+	// Metric sanitization type.
+	MetricSanitization *MetricSanitizationType `yaml:"sanitization"`
```
xichen2020: nit: just call this `Sanitization`. This is a metric configuration object so don't think we need another `metric` in the name.

instrument/config.go
```diff
@@ -94,10 +97,16 @@ func (mc *MetricsConfiguration) NewRootScopeReporter(
 		}
 	}
 
+	var sanitizeOpts *tally.SanitizeOptions
+	if mc.MetricSanitization != nil {
+		sanitizeOpts = mc.MetricSanitization.Options()
```
xichen2020: nit: rename `Options` to `NewOptions` to be consistent with other config-related constructors.

instrument/sanitize.go
```diff
@@ -0,0 +1,90 @@
...
+}
+
+// UnmarshalYAML unmarshals a MetricSanitizationType into a valid type from string.
+func (t *MetricSanitizationType) UnmarshalYAML(unmarshal func(interface{}) error) error {
+	var str string
+	if err := unmarshal(&str); err != nil {
+		return err
+	}
+	if str == "" {
+		*t = NoMetricSanitization
```
xichen2020: Might make sense to add a `defaultMetricSanitization` to the list of sanitization consts and use it here.

instrument/sanitize.go
```diff
@@ -37,6 +37,9 @@ const (
 
 	// M3MetricSanitization performs M3 metric sanitization.
 	M3MetricSanitization
+
+	// DefaultMetricSanitization is the default extended metrics level.
+	DefaultMetricSanitization = NoMetricSanitization
```
xichen2020: nit: this doesn't need to be exported.

PR #89, opened 2017-08-20
Add build version reporter

prateek:
- Adds a new interface to report runtime version and githash (expected at compile)
- Adds leaktests glide dependency 

/cc @xichen2020 @robskillington @jeromefroe


instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+	"sync/atomic"
+	"time"
+)
+
+var (
+	// Githash is the githash associated with this build. Overridden using ldflags
+	// at compile time. Example:
+	// $ go build -ldflags "-X github.com/m3db/m3x/instrument.Githash=abcdef" ...
+	// Adapted from: https://www.atatus.com/blog/golang-auto-build-versioning/
+	Githash = "unknown"
```
jeromefroe: nit: maybe `Revision` would be more agnostic of our version control system?

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+
+	// GoVersion is the current runtime version
+	goVersion = runtime.Version()
```
jeromefroe: nit: perhaps make `goVersion` public? Seems odd that `Githash` would be but it wouldn't.

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+)
+
+var (
+	// Githash is the githash associated with this build. Overridden using ldflags
+	// at compile time. Example:
+	// $ go build -ldflags "-X github.com/m3db/m3x/instrument.Githash=abcdef" ...
+	// Adapted from: https://www.atatus.com/blog/golang-auto-build-versioning/
+	Githash = "unknown"
+
+	// GoVersion is the current runtime version
```
jeromefroe: nit: add a period to the end of these comments?

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+		case <-ticker.C:
+			gauge.Update(1.0)
+		case <-v.closeCh:
+			return
+		}
+	}
+}
+
+func (v *versionReporter) Close() error {
+	if active := atomic.LoadInt64(&v.active); active == 0 {
```
jeromefroe: I think you could get a panic here if `Close` is called more than once concurrently because `v.closeCh` will be closed more than once. Might be better to replace this with a lock since it will likely only be called once?

jeromefroe:
This is awesome! Would be really great to log this info once any of our binaries start, would make debugging a lot easier.

instrument/version.go
prateek: sure thing

instrument/version.go
prateek: I'd prefer to keep it un-exported to guarantee the value is set as `runtime.Version`.

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+	// $ go build -ldflags "-X github.com/m3db/m3x/instrument.Githash=abcdef" ...
+	// Adapted from: https://www.atatus.com/blog/golang-auto-build-versioning/
+	Githash = "unknown"
+
+	// GoVersion is the current runtime version
+	goVersion = runtime.Version()
+
+	// metricName is the emitted metric's name
+	metricName = "build-information"
+)
```
prateek: Sure. 

Of the ones listed:
- I don't think `Version` makes sense for us. 
- Dunno if `BuildUser` is available via some uDeploy env flag.
- I'll pull `BuildDate` too

instrument/version.go
prateek: sue thing

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+	}
+	atomic.StoreInt64(&v.active, 1)
+	v.closeCh = make(chan struct{}, 1)
+	v.doneCh = make(chan struct{}, 1)
+	go v.report()
+	return nil
+}
+
+func (v *versionReporter) report() {
+	v.opts.Logger().Infof("Go runtime version: %s", goVersion)
```
prateek: That's fair. Will do

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+	v.opts.Logger().Infof("Build githash: %s", Githash)
+
+	scope := v.opts.MetricsScope().Tagged(map[string]string{
+		"githash":    Githash,
+		"go.version": goVersion,
```
prateek: Sure thing

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+func NewVersionReporter(
+	opts Options,
+) VersionReporter {
+	return &versionReporter{
+		opts: opts,
+	}
+}
+
+func (v *versionReporter) Start() error {
+	if active := atomic.LoadInt64(&v.active); active != 0 {
```
prateek: Good catch, will fix

instrument/version.go
```diff
@@ -0,0 +1,111 @@
...
+		select {
+		case <-ticker.C:
+			gauge.Update(1.0)
+		case <-v.closeCh:
+			return
+		}
+	}
+}
+
+func (v *versionReporter) Close() error {
```
prateek: It won't panic once the CAS fix is put in (from the comment above).

instrument/version.go
prateek: done

instrument/version.go
prateek: Didn't find `BuildUser` on the uBuild EngDocs. Added others.

instrument/version.go
prateek: done

instrument/version.go
prateek: Created the function, and take in another flag `LogBuildInformationAtStartup` checked at `init()` time, this way we can control whether a build logs this information at compile time. W/o cluttering the main for each binary.

instrument/version.go
prateek: done

instrument/version.go
prateek: done

instrument/version.go
prateek: done

instrument/build.go
```diff
@@ -0,0 +1,133 @@
...
+	return &buildReporter{
+		opts: opts,
+	}
+}
+
+func (v *buildReporter) Start() error {
+	if swapped := atomic.CompareAndSwapInt64(&v.active, 0, 1); !swapped {
+		return errAlreadyStarted
+	}
+	v.closeCh = make(chan struct{}, 1)
```
jeromefroe: nit: probably don't need to buffer `closeCh` since we don't actually put any values onto it.

instrument/build.go
```diff
@@ -0,0 +1,133 @@
...
+		case <-ticker.C:
+			gauge.Update(1.0)
+		case <-v.closeCh:
+			return
+		}
+	}
+}
+
+func (v *buildReporter) Close() error {
+	if active := atomic.LoadInt64(&v.active); active == 0 {
```
jeromefroe: nit: admittedly an extreme edge case, but multiple concurrent calls to `Close` could cause a panic, if `v.closeCh` is closed more than once. Perhaps we want to use a mutex or state that it cannot be called concurrently?

instrument/build.go
prateek: will do

instrument/build.go
prateek: fair point, changed from atomic to locks to prevent that.

instrument/build.go
prateek: done

instrument/build.go
prateek: done

instrument/build.go
```diff
@@ -0,0 +1,140 @@
...
+	metricName = "build-information"
+)
+
+var (
+	errAlreadyStarted = errors.New("reporter already started")
+	errNotStarted     = errors.New("reporter not started")
+)
+
+// LogBuildInformation logs the build information to the provided logger.
+func LogBuildInformation() {
```
jeromefroe: super duper nit: maybe `LogBuildInfo` instead of `LogBuildInformation` to match `LogBuildInfoAtStartup`

PR #91, opened 2017-08-22
[time] Add a Duration method to Range and a Count method to Unit

jeromefroe:
This PR adds a `Duration` method to the `Range` struct in the `time` package which returns the duration of the range a `Count` method to `Unit` which returns the number of units contained within a duration.

cc @xichen2020

### DIFF
--- a/time/range.go
+++ b/time/range.go
@@ -61,6 +61,11 @@ func (r Range) Overlaps(other Range) bool {
 	return r.End.After(other.Start) && r.Start.Before(other.End)
 }
 
+// Duration returns the duration of the range.
+func (r Range) Duration() time.Duration {
+	return r.End.Sub(r.Start)
+}
+
 // Intersect calculates the intersection of the receiver range against the
 // provided argument range iff there is an overlap between the two. It also
 // returns a bool indicating if there was a valid intersection.
--- a/time/unit.go
+++ b/time/unit.go
@@ -57,11 +57,17 @@ func (tu Unit) Value() (time.Duration, error) {
 }
 
 // Count returns the number of units contained within the duration.
-func (tu Unit) Count(d time.Duration) (int, error) {
-	if dur, found := unitsToDuration[tu]; found {
-		return int(d / dur), nil
+func (tu Unit) Count(d time.Duration) int {
+	if d < 0 {
+		return -1
 	}
-	return 0, errUnrecognizedTimeUnit
+
+	if dur, found := unitsToDuration[tu]; found {
+		return int(d / dur)
+	}
+
+	// Invalid unit.
+	return -1
 }
 
 // IsValid returns whether the given time unit is valid / supported.
--- a/time/unit.go
+++ b/time/unit.go
@@ -43,6 +43,7 @@ var (
 	errConvertDurationToUnit = errors.New("unable to convert from duration to time unit")
 	errConvertUnitToDuration = errors.New("unable to convert from time unit to duration")
 	errMaxUnitForDuration    = errors.New("unable to determine the maximum unit for duration")
+	errNegativeDuraton       = errors.New("duration cannot be negative")
 )
 
 // Unit represents a time unit.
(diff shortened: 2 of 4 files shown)


time/range.go
```diff
@@ -61,6 +61,11 @@ func (r Range) Overlaps(other Range) bool {
 	return r.End.After(other.Start) && r.Start.Before(other.End)
 }
 
+// Duration returns the duration of the range.
+func (r Range) Duration() time.Duration {
+	return r.End.Sub(r.Start)
```
prateek: Do you need to consider negative durations at all here?

time/unit.go
```diff
@@ -56,6 +56,14 @@ func (tu Unit) Value() (time.Duration, error) {
 	return 0, errUnrecognizedTimeUnit
 }
 
+// Count returns the number of units contained within the duration.
+func (tu Unit) Count(d time.Duration) (int, error) {
```
prateek: nit: make this `int64` instead of `int`

time/unit.go
```diff
@@ -56,6 +56,14 @@ func (tu Unit) Value() (time.Duration, error) {
...
+	if dur, found := unitsToDuration[tu]; found {
+		return int(d / dur), nil
```
prateek: same point as above about negative counts. Might be better to guarantee `>=0` in the contract.

time/unit.go
```diff
@@ -56,6 +56,14 @@ func (tu Unit) Value() (time.Duration, error) {
...
+	}
+	return 0, errUnrecognizedTimeUnit
```
prateek: Feels weird to have a method on `Unit` return an error on unrecognised units. Maybe push this on the user, where they have to guarantee the Unit IsValid before calling this, can do an internal check in this method and return `-1` if invalid.

time/range.go
jeromefroe: The other tests all take as invariant that `Start` <= `End` so I didn't want to special case this one. Also, this would actually be fine since `time.Duration` is a type alias for `int64` and can be negative.

time/unit.go
jeromefroe: Good point, I'll require that `d` is greater than 0.

time/unit.go
jeromefroe: Hmm, what makes you prefer `int64` to `int` here?

time/unit.go
jeromefroe: Sounds good to me.

time/range.go
prateek: Yeah was more worried about the semantic meaning of a range with a negative duration, doesn't matter I guess.

time/unit.go
prateek: `time.Duration` is aliased to int64, so dividing here leaves us with `int64`. No reason to loose range of values by down casting

time/unit.go
```diff
@@ -56,6 +56,20 @@ func (tu Unit) Value() (time.Duration, error) {
...
 }
 
+// Count returns the number of units contained within the duration.
+func (tu Unit) Count(d time.Duration) int {
+	if d < 0 {
+		return -1
+	}
+
+	if dur, found := unitsToDuration[tu]; found {
+		return int(d / dur)
```
prateek: no need to downcast here, return `int64`.

time/unit.go
```diff
@@ -56,6 +56,20 @@ func (tu Unit) Value() (time.Duration, error) {
 	return 0, errUnrecognizedTimeUnit
 }
 
+// Count returns the number of units contained within the duration.
+func (tu Unit) Count(d time.Duration) int {
+	if d < 0 {
+		return -1
```
xichen2020: Hm why not return an (int, error)? Returning -1 seems weird.

jeromefroe: I did that originally, but as Prateek pointed out it's a little weird for a function as simple as Count to return an error. It's kind of like `bytes.Index`, is the value not being present really worth an error? I don't feel too strongly though, wouldn't mind going back to the original way.

time/unit.go
jeromefroe: To be honest, I feel like `int` is the more appropriate type to return. Most functions in the standard library return an `int` as opposed to an `int64` so I feel that it would be more idiomatic.

time/unit.go
xichen2020: Well I think it's unconventional for `Count` to return -1 when the input is invalid. Why not have a `Count` return `(int, error)` and a `MustCount` return `int` and panics if the input is invalid? This way the caller can always use `MustCount` if input is known to be valid. This is the same approach adopted in other repos as well as in many golang packages (`Compile` and `MustCompile` in the `regexp` package)

jeromefroe: Sounds good to me, in fact I like the `Count` and `MustCount` approach better than just `Count` with either signature.

time/unit.go
jeromefroe: Must have missed this comment, I feel it would be more idiomatic to return an `int` instead of `int64`. For the majority of use cases I imagine it will be a noop anyhow since it'll be run on a 64 bit platform anyway.

prateek: Sounds good

PR #92, opened 2017-08-28
[godoc] And package level documentation and examples

jeromefroe:
This PR adds package level documentation for the packages in `m3x` and adds examples for some of the packages. The goal of this PR is to improve the documentation for the `m3x` package.

cc @xichen2020 @prateek @cw9


clock/types.go
```diff
@@ -18,21 +18,22 @@
...
 
-// NowFn is the function supplied to determine "now"
+// NowFn is the function supplied to determine "now".
 type NowFn func() time.Time
 
-// Options represents the options for the clock
+// Options represents the options for the clock.
 type Options interface {
-	// SetNowFn sets the nowFn
+	// SetNowFn sets the nowFn.
```
prateek: nit: // sets the *NowFn*

clock/types.go
```diff
@@ -18,21 +18,22 @@
...
 	SetNowFn(value NowFn) Options
 
-	// NowFn returns the nowFn
+	// NowFn returns the nowFn.
```
prateek: nit: // sets the *NowFn*.

close/close.go
```diff
@@ -18,6 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+// Package xclose provides utilities for closing resources.
```
prateek: this is something weird I've noticed about this repo. Some packages are prefixed with 'x', and some aren't. Do you know what the deal is?

config/example_test.go
```diff
@@ -0,0 +1,38 @@
...
+)
+
+type config struct {
+	ListenAddress string `yaml:"listenAddress" validate:"nonzero"`
+}
+
+func ExampleLoadFile() {
+	cfg := &config{}
+	if err := xconfig.LoadFile(cfg, "/path/to/config/file"); err != nil {
+		log.Fatal(err)
```
prateek: Examples without `// Output: ` are compiled iirc. Could you confirm this is being run in the test suite?

errors/example_test.go
```diff
@@ -0,0 +1,44 @@
...
+
+	for i := 0; i < 5; i++ {
+		_, err := http.Get("http://example.com/")
+		if err != nil {
+			// Add returns a new MultiError.
+			mutliErr = mutliErr.Add(err)
+		}
+	}
+
+	if err := mutliErr.FinalError(); err != nil {
```
prateek: same as above: confirm this is actually run.

hash/jump/example_test.go
```diff
@@ -0,0 +1,47 @@
...
+	// Create hash of the key using whatever hash function you wish.
+	if _, err := hasher.Write(key); err != nil {
+		log.Fatal(err)
+	}
+	keyHash := hasher.Sum64()
+
+	// Get which bucket the key is assigned to.
+	bucket := jump.Hash(keyHash, numBuckets)
+
+	log.Printf("key '%s' is assigned to bucket %d\n", string(key), bucket)
```
prateek: same as above, probably need an `// Output: ` block here

net/example_test.go
```diff
@@ -0,0 +1,60 @@
...
+	}
+	defer l.Close()
+
+	// Start accepting incoming connections.
+	var (
+		opts          = xretry.NewOptions()
+		connCh, errCh = xnet.StartAcceptLoop(l, opts)
+	)
+
+	for {
```
prateek: same as above: does this run in the test suite?

pool/example_test.go
```diff
@@ -0,0 +1,54 @@
...
+	a, b, c int64
+}
+
+func (o *exampleObject) reset() {
+	o.a = 0
+	o.b = 0
+	o.c = 0
+}
+
+func ExampleObjectPool() {
```
prateek: same as above, need an `// Output` block

retry/example_test.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"context"
+	"log"
+	"net/http"
+
+	"github.com/m3db/m3x/retry"
+)
+
+func ExampleRetrier() {
```
prateek: same as above, need an `// Output` block

close/close.go
jeromefroe: I have no idea, but it did irk me as I was going through this since there didn't seem to be any consistent thread to it that I could pick up on. Might be worth agreeing on a convention and then updating each package one by one, as necessary. I think if we split up the work it shouldn't be too bad, just some find and replaces in the other m3db repos and statsdex.

config/example_test.go
jeromefroe: If you include `// Output:` the test is both compiled an run as part of the test suite, if you don't include it the test is only compiled. I think only compiling it makes sense in this case because otherwise we would have to write a temporary config file to ensure the call to `config.LoadFile` works properly. I thought about it in this case but I wasn't aware of a way that I could the necessary setup code but have it hidden in the example, and I was concerned that including it in the example would make it more cumbersome and less clear. Let me know what you think though, I'm not certainly not fully convinced that this is the right approach.

errors/example_test.go
jeromefroe: Following up on the previous comment on `// Output:`, I think including it here makes sense. Will update.

hash/jump/example_test.go
jeromefroe: Yep, this certainly should include an `// Output:` comment. Will update.

net/example_test.go
jeromefroe: I think in this case we want to just compile the test since otherwise we would have to introduce some kind of timeout to close the server which may make the example less clear.

pool/example_test.go
jeromefroe: This certainly should be run. Will update.

retry/example_test.go
jeromefroe: Yep, no reason not run this one, will update.

server/example_test.go
```diff
@@ -0,0 +1,50 @@
...
+
+	"github.com/m3db/m3x/server"
+)
+
+type simpleHandler struct{}
+
+func (h *simpleHandler) Handle(conn net.Conn) { conn.Close() }
+func (h *simpleHandler) Close()               {}
+
+func ExampleServer() {
```
jeromefroe: @prateek I think I will leave this one just as compiled since we would have to introduce some arbitrary mechanism to close the server (like in the `ListenAndServe` example which might make the example less clear.

sync/example_test.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"log"
+	"net/http"
+	"sync"
+
+	"github.com/m3db/m3x/sync"
+)
+
+func ExampleWorkerPool() {
```
jeromefroe: @prateek Will run this one though, no reason not to in my opinion.

close/close.go
prateek: Sounds good, how about changing the names in this PR while you're here. 

We can split the work to update repos once this lands.

config/example_test.go
prateek: I tried writing up an example for this (below), and it looks like a unit test to me. IMO your approach is a good middle ground. Plus, go's stdlib seems to follow the same ~ e.g. https://golang.org/pkg/io/ioutil/#example_ReadFile


```
func ExampleLoadFile() {
  exampleData := byte[](`listenAddress: "0.0.0.0:12321"`)
  if err := ioutil.WriteFile("testFile", exampleData, ...) ; err != nil {
    log.Fatalf("unable to write example file: %v", err)
  }
  var cfg config
  if err := xconfig.LoadFile(&cfg, "exampleFile"); err != nil {
     log.Fatalf("unable to load file: %v", err)
  }
  fmt.Println("listenAddress: %s", cfg.ListenAddress)
  // Output: ...
}
```

close/close.go
jeromefroe: Sounds good to me, I'll probably hold off until Friday when Xi gets back, just so I can make sure I didn't miss any pattern to the naming.

config/example_test.go
jeromefroe: Ahh, nice. Seems the `ReadFile` example is a perfect model of what we want to do here. They include a `testdata` directory which contains the file itself. I think I'll add one here as well so we can run this as a unit test, but still hide the initial setup of actually writing the test file.

errors/example_test.go
jeromefroe: Updated this, GIthub just didn't notice because this particular line didn't change.

pool/example_test.go
jeromefroe: Updated this as well, GIthub didn't notice because this particular line didn't change.

retry/example_test.go
jeromefroe: Same here as well, updated this but GIthub just didn't notice because this particular line didn't change.

close/close.go
jeromefroe: FWIW, I just took a look at the packages under the `x` repo for `golang.org` and they all use the normal package not name, not "x<name>": https://godoc.org/-/subrepo

prateek: No need to block this PR, opened https://github.com/m3db/m3x/issues/95 to track the issue. Feel free to land this w/o the package naming fix.

prateek: I definitely prefer us not using the `x` prefix, unless we rename the package too.

config/example_test.go
prateek: sounds good!

config/testdata/conf.yaml
```diff
@@ -0,0 +1 @@
+listenAddress: "0.0.0.0:8392"
```
prateek: nit: newline

errors/example_test.go
prateek: nit: Have the `Output` be driven off an error message specified in one of the `err`s

net/example_test.go
prateek: Sounds good

pool/example_test.go
prateek: nit: could you drive the output off the retrieved object, something like -

```
fmt.Printf("Retrieved struct should have default values: %+v", o) 
// Output: ... 
```

retry/example_test.go
prateek: nit: same as above, drive the output of the example somehow - either capturing the closed context case, or counting the number of tries.

server/example_test.go
prateek: Sounds good

prateek:
@jeromefroe mainly looks good. Just a few nits here and there. 

Could you also add a go-doc badge (e.g. https://github.com/kubernetes/kubernetes/blame/master/README.md#L77)

jeromefroe:
@prateek looks like we have a Godoc badge already but the link is broken because we don't have a top level `doc.go` file. This diff adds one though so it should work once this lands.

pool/example_test.go
```diff
@@ -0,0 +1,62 @@
...
+	fmt.Printf("Retrieved struct should have default values: %+v", o)
+
+	// Use the exampleObject.
+	_ = o
+
+	// Reset the exampleObject and return it to the pool.
+	o.reset()
+	p.Put(o)
+
+	// Output: Retrieved struct should have default values: &{a:0 b:0 c:0}
```
prateek: nit: could you move to be right after the `fmt.Printf`

retry/example_test.go
```diff
@@ -0,0 +1,65 @@
...
+			return nil
+		}
+		return errors.New("test")
+	}
+
+	if err := retrier.AttemptWhile(continueFn, fn); err != nil {
+		log.Fatal(err)
+	}
+
+	// Output: Attempt 3 succeeded
```
prateek: nit: move to be by the `Printf`

PR #93, opened 2017-09-03
[instrument] Add zap logger to options interface

jeromefroe:
This PR adds a `zap` logger to the `instrument.Options` interface. `zap` offers a performant logger which is actively maintained. In addition, it provides a number of other features include log sampling and the ability to dynamically change the log level. By adding `zap` as a new method on the `Options` interface we can migrate current packages piecemeal as the need arises.

### DIFF
--- a/instrument/options.go
+++ b/instrument/options.go
@@ -45,10 +45,9 @@ type options struct {
 // NewOptions creates new instrument options.
 func NewOptions() Options {
 	logger := log.NewLevelLogger(log.SimpleLogger, log.LevelInfo)
-	zap := zap.L()
 	return &options{
 		logger:         logger,
-		zap:            zap,
+		zap:            zap.L(),
 		scope:          tally.NoopScope,
 		samplingRate:   defaultSamplingRate,
 		reportInterval: defaultReportingInterval,
--- a/instrument/options.go
+++ b/instrument/options.go
@@ -24,7 +24,9 @@ import (
 	"time"
 
 	"github.com/m3db/m3x/log"
+
 	"github.com/uber-go/tally"
+	"go.uber.org/zap"
 )
 
 const (
@@ -34,6 +36,7 @@ const (
 
 type options struct {
 	logger         log.Logger
+	zap            *zap.Logger
 	scope          tally.Scope
 	samplingRate   float64
 	reportInterval time.Duration
@@ -42,8 +45,10 @@ type options struct {
 // NewOptions creates new instrument options.
 func NewOptions() Options {
 	logger := log.NewLevelLogger(log.SimpleLogger, log.LevelInfo)
+	zap := zap.L()
 	return &options{
 		logger:         logger,
+		zap:            zap,
 		scope:          tally.NoopScope,
 		samplingRate:   defaultSamplingRate,
 		reportInterval: defaultReportingInterval,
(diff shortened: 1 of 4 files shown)


instrument/options.go
```diff
@@ -35,6 +37,7 @@ const (
 
 type options struct {
 	logger         xlog.Logger
+	zap            *zap.Logger
```
prateek: Instead of this, would it be better to write a `xlog.Logger` implementation that wrapped `zap.Logger`?

instrument/options.go
```diff
@@ -26,6 +26,8 @@ import (
 	"github.com/m3db/m3x/log"
 
 	"github.com/uber-go/tally"
+
```
prateek: nit: omit this newline

instrument/options.go
```diff
@@ -45,6 +48,7 @@ func NewOptions() Options {
 	logger := xlog.NewLevelLogger(xlog.SimpleLogger, xlog.LogLevelInfo)
 	return &options{
 		logger:         logger,
+		zap:            zap.NewNop(),
```
prateek: maybe a better default would be `zap.Production` or `zap.L`?

instrument/options.go
prateek: Long term we should straight up replace `xlog.Logger` with `zap.Logger` - for the reasons you list, and also because Zap uses a more performant `LogField` implementation (can specify which type serialiser to use, avoid allocs unless necessary).

That might be too pervasive to start tho, starting with the wrapped impl would be easier.

prateek: Thought about this some more, exposing the `zap.Logger` type (the way you're doing it), is a better middle ground. 

Reasoning (for posterity): this way allows downstream packages (m3db/m3aggregator/etc) to get off our logger implementation without breaking other consumers. Will make the process to remove our own implementation of log easier. Wrapping the Zap impl behind our interface is throw away work in the long term.

instrument/options.go
prateek: Mainly because the xlog logger we provide by default writes to stdout.

prateek:
@jeromefroe mainly looks good, just change the default provided in the ctor and rebase. 

Other thoughts: 
- provide utility methods for common construction patterns 
- log rotation setup (e.g. from Zap [FAQ](https://github.com/uber-go/zap/blob/master/FAQ.md#does-zap-support-log-rotation))

instrument/options.go
```diff
@@ -44,6 +47,7 @@ func NewOptions() Options {
 	logger := log.NewLevelLogger(log.SimpleLogger, log.LevelInfo)
 	return &options{
 		logger:         logger,
+		zap:            zap.L(),
```
jeromefroe: Thought about using `zap.Production` here, but it returns an error and I didn't see any clean way to handle that error. We could change the type signature to bubble the error up but it's very unintuitive that a call to `NewOptions` would return an error, especially since in practice it shouldn't ever return an error. Alternatively, we could panic if there was an error but that has a code smell to it. So I decided to just go with `zap.L()` as the default here.

jeromefroe:
@prateek mind taking another pass when you get a chance? I admittedly don't have much experience with `zap` so I'm not entirely sure how to add support for common construction patterns and log rotation cleanly? Perhaps we start with this as the bare bones and when we start to see common code being duplicated in multiple places then augment it here?

instrument/options.go
prateek: Sounds good

prateek:
@jeromefroe sounds good

PR #97, opened 2017-09-10
[travis] Add go 1.8 and 1.9 to travis tests

jeromefroe:
This PR adds Go 1.8 and 1.9 to the versions of Go we test against in Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,8 @@
 language: go
 go:
   - 1.7
+  - 1.8
+  - 1.9
 install:
   - go get -v github.com/Masterminds/glide
   - glide install
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,11 +1,9 @@
 language: go
 go:
-  - 1.7
-  - 1.8
+  - 1.7.6
+  - 1.8.3
   - 1.9
-install:
-  - go get -v github.com/Masterminds/glide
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -1,6 +1,8 @@
 language: go
 go:
   - 1.7
+  - 1.8
+  - 1.9
```
prateek: same as other PR, change to latest point releases

.travis.yml
```diff
@@ -1,6 +1,8 @@
...
 install:
   - go get -v github.com/Masterminds/glide
   - glide install
```
prateek: while you're here, could you change these to use the appropriate targets from the Makefile

prateek:
@jeromefroe I wonder if we should take this opportunity to refactor all makefiles a bit too - all repos seem to have a lot of targets in common (`lint`/`metalint`/install-.../etc.) Could migrate them to `ci-scripts/common.mk`, thoughts?

jeromefroe:
Yea, probably wouldn't be a bad idea, there's a fair amount of duplication between the repos in `lint`, `metalint`, etc.

prateek:
@jeromefroe cool, i'll take a pass at extracting common bits into ci-scripts and do an example repo. we can split up the repos after.

Issue #95, opened 2017-09-06
Make package and folder names identical


prateek:
We have some packages in the repo who's names don't match the folders they reside in. E.g. [m3x/log](https://github.com/m3db/m3x/blob/master/log/logger.go#L21) uses `xlog` as package name, and `log` as folder name. Simultaneously, we have packages like [instrument](https://github.com/m3db/m3x/blob/master/instrument/types.go#L21), which have package name == folder name. 

This leads to a bad user experience, and causes issues when using `goimports` to resolve imports (e.g. have to manually import `xlog` most of the time because it'll pick the stdlib `log` usually). 

We should either make all the packages prefix `x`, or none.

jeromefroe:
All the open source resources have been updated, so I'm going to close this out now.

PR #98, opened 2017-09-13
[refactor] Use consistent naming approach for packages

jeromefroe:
This package addresses #95 and ensures that we use a consistent naming approach for our packages by removing the `x` from all packages which begin with one. Will take a stab at updating all packages that depend on `m3x` when this lands.


prateek:
LGTM

PR #99, opened 2017-09-16
[glide] Only include direct dependencies and use semantic versions where available

jeromefroe:
This PR updates our approach to vendoring dependencies with Glide. Previously, we would almost exclusively use SHA's as opposed to versioned releases and we would vendor transitive dependencies in our `glide.yaml` file in addition to direct dependencies. I think each of these techniques has flaws given the current state of package management in Go however. 

The problem with using SHA's is that Glide has no flexibility in choosing which version of a package to vendor so all vendored packages have to agree on the exact same version. This was highlighted recently when we tried to update the version of `tally` in statsdex. We wanted to use a new feature in a minor release of `tally` but in order to vendor it in statsdex we first had to update each `m3db` repo that statsdex depended on since they were pinned to an older version and did not support minor version changes. By allowing minor version changes between packages we can allow packages to update when they want to pull in new features or updates, and still rely on the API stability of the major version so that packages that do not need these new features will continue to work as expected.

The second way I think we can improve how we use Glide is to not vendor transitive dependencies in our `glide.yaml` files. When we add transitive dependencies to `glide.yaml` there is no way to distinguish them as such. Consequently, when updating dependencies we have to remember which ones depend on others so that we can update them accordingly. Instead, given that package mangement has matured greatly in the Go ecosystem and there are [official tools in development](https://github.com/golang/dep) we should instead allow packages to manage their own dependencies and not introduce them into `glide.yaml`. Under this approach we still have reproducible builds since `glide.lock` tracks the exact version of a package being used at any given time, and `glide install` will only install those versions. Furthermore, Glide offers the ability to update only a subset of packages (`glide update -v <package>`) so that when we do update dependencies we can be selective about which ones we choose.

Anyway, that's a lot of rambling, let me know what you think @xichen2020 @prateek @cw9

### DIFF
--- a/glide.lock
+++ b/glide.lock
@@ -1,5 +1,5 @@
-hash: 50a0b59635ffd9cd3f6bd65a7f6a415f0f3a9a82d4f40b997585e04a4bcf2e86
-updated: 2017-08-20T15:45:39.86828054-04:00
+hash: 2bd54f77e2cac4a0b501cc4032ea9e21a06002248a54cb4a11f33e74cb2410a3
+updated: 2017-09-16T09:13:52.427152982-04:00
 imports:
 - name: github.com/apache/thrift
   version: 9549b25c77587b29be4e0b5c258221a4ed85d37a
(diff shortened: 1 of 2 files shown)


cw9:
rebase from master?

PR #100, opened 2017-09-27
Change default level to simple and move FD metrics to moderate level

xichen2020:
cc @jeromefroe @cw9 

This PR introduces a `Moderate` level that's between the `Simple` and `Detailed` and moves file descriptor reporting to the `Moderate` level for improved flexibility.

### DIFF
--- a/instrument/extended.go
+++ b/instrument/extended.go
@@ -43,8 +43,11 @@ const (
 	// SimpleExtendedMetrics describes just a simple level of extended metrics:
 	// - number of active goroutines
 	// - number of configured gomaxprocs
-	// - number of file descriptors
 	SimpleExtendedMetrics
+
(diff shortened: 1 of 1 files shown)


instrument/extended.go
```diff
@@ -43,9 +43,12 @@ const (
 	// SimpleExtendedMetrics describes just a simple level of extended metrics:
 	// - number of active goroutines
 	// - number of configured gomaxprocs
-	// - number of file descriptors
 	SimpleExtendedMetrics
 
+	// ModerateExtendedMetrics describes a moderately verbose level of extended metrics:
+	// - number of file descriptors
```
jeromefroe: nit: shouldn't this include:
```
// - number of active goroutines
// - number of configured gomaxprocs
```

xichen2020: Yup good catch.

PR #101, opened 2017-09-27
Add mumur3 128-bit as a more efficient alternative for hashing ids

xichen2020:
cc @jeromefroe @cw9 @prateek 

This PR adds murmur3 as a more efficient and equally good collision-wise alternative to generate 128-bit hashes of ids.

```
BenchmarkMD5ShortID-8               	10000000	       151 ns/op
BenchmarkMD5MediumID-8              	 3000000	       421 ns/op
BenchmarkMD5LongID-8                	  200000	     11567 ns/op
BenchmarkMurmur3Hash128ShortID-8    	100000000	        24.4 ns/op
BenchmarkMurmur3Hash128MediumID-8   	30000000	        50.8 ns/op
BenchmarkMurmur3Hash128LongID-8     	 1000000	      1232 ns/op
```

### DIFF
--- a/id/hash_bench_test.go
+++ b/id/hash_bench_test.go
@@ -36,40 +36,66 @@ var (
 	testShortID, testMediumID, testLongID []byte
 )
 
-func BenchmarkMD5ShortID(b *testing.B) {
+func benchmarkMD5ShortID(b *testing.B) {
 	for i := 0; i < b.N; i++ {
 		HashFn(testShortID)
 	}
 }
 
-func BenchmarkMD5MediumID(b *testing.B) {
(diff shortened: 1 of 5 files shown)


id/hash_bench_test.go
```diff
@@ -0,0 +1,88 @@
...
+
+func generateTestID(s *rand.Rand, size int) []byte {
+	id := make([]byte, size)
+	for i := 0; i < size; i++ {
+		id[i] = byte(s.Int63() % 256)
+	}
+	return id
+}
+
+func init() {
```
jeromefroe: Just curious, why are we including these functions in the `init` function?

PR #106, opened 2017-10-17
Add instrumented worker pool

prateek:
Ports internal instrumented worker pool to m3x

/cc @xichen2020 @martin-mao @robskillington  @jeromefroe

### DIFF
--- a/sync/instrumented_worker_pool..go
+++ b/sync/instrumented_worker_pool..go
@@ -60,10 +60,7 @@ func (p *instrumentedWorkerPool) Close() error {
 		return errAlreadyClosed
 	}
 
-	if err := p.pool.Close(); err != nil {
-		return err
-	}
-	return nil
+	return p.pool.Close()
 }
 
 func (p *instrumentedWorkerPool) metricLoop() {
--- a/sync/instrumented_worker_pool..go
+++ b/sync/instrumented_worker_pool..go
@@ -26,6 +26,7 @@ import (
 	"time"
 
 	"github.com/m3db/m3x/instrument"
+
 	"github.com/uber-go/tally"
 )
 
--- a/sync/instrumented_worker_pool..go
+++ b/sync/instrumented_worker_pool..go
@@ -1,0 +1,161 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+//
+// The above copyright notice and this permission notice shall be included in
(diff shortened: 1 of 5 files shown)


sync/instrumented_worker_pool_test.go
```diff
@@ -0,0 +1,180 @@
...
+
+	errors := checkMetricsExtended(wantG, wantC, s)
+	if len(errors) == 0 {
+		return
+	}
+	assert.FailNow(t, strings.Join(errors, "\n"))
+}
+
+func TestInstrumentedWorkerPool(t *testing.T) {
+	scope := tally.NewTestScope("", nil)
```
prateek: add a leaktest check here

sync/instrumented_worker_pool..go
```diff
@@ -0,0 +1,162 @@
...
+	}
+}
+
+func (p *instrumentedWorkerPool) Init() {
+	p.pool.Init()
+	p.metrics.incIdle(int64(p.size))
+	go p.metricLoop()
+}
+
+func (p *instrumentedWorkerPool) Close() error {
```
prateek: add a test case for this method

sync/instrumented_worker_pool_test.go
prateek: done

sync/instrumented_worker_pool..go
prateek: done

sync/worker_pool.go
```diff
@@ -94,3 +81,17 @@ func (p *workerPool) GoWithTimeout(work Work, timeout time.Duration) bool {
 		return false
 	}
 }
+
+func (p *workerPool) Close() error {
+	if len(p.workCh) < cap(p.workCh) {
+		return errAllRoutinesNotFinished
+	}
+
+	set := atomic.CompareAndSwapInt64(&p.closed, 0, 1)
```
xichen2020: I think there is a chance in between the check above and the atomic swap here that someone may call `p.Go`, so if you close the channel below before that call returns, the program will panic due to enqueuing to a closed channel.

jeromefroe: +1

I think we also need to change the semantics of the `select` statements if we add a `Close` method. Otherwise, once the channel is closed, `token := <-p.workCh` will return immediately, and we wouldn't know that it is because the channel is closed and not because a token was returned to the pool. So perhaps we should change the call to `token, ok := <-p.workCh` and check that `ok` is true?

sync/instrumented_worker_pool..go
```diff
@@ -0,0 +1,158 @@
...
+type workerPoolMetrics struct {
+	timeouts    tally.Counter
+	unavailable tally.Counter
+	run         tally.Counter
+	idle        tally.Gauge
+	busy        tally.Gauge
+	waiting     tally.Gauge
+	total       tally.Gauge
+
+	numIdle    int64
```
jeromefroe: nit: same comment here about the atomics

robskillington:
Closing for now, unless there is a real need this is both considerably more expensive than our existing worker pool and has plenty of gotchas.

Won't delete the branch if you want to revive later.

PR #107, opened 2017-10-19
Add hostid config package for resolving a host ID

robskillington:
This is a port as it might be useful for other projects that resolve the host ID in a configuration driven manner.

As you suggested @prateek.

### DIFF
--- a/config/hostid/hostid.go
+++ b/config/hostid/hostid.go
@@ -18,6 +18,8 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
+// Package hostid provides a configuration struct for resolving
+// a host ID from YAML.
 package hostid
 
 import (
--- a/config/hostid/hostid.go
+++ b/config/hostid/hostid.go
@@ -1,0 +1,81 @@
+// Copyright (c) 2017 Uber Technologies, Inc.
+//
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


config/hostid/hostid.go
```diff
@@ -0,0 +1,81 @@
...
+)
+
+// Resolver is a type of host ID resolver
+type Resolver string
+
+const (
+	// HostnameResolver resolves host using the hostname returned by OS
+	HostnameResolver Resolver = "hostname"
+	// ConfigResolver resolves host using a value provided in config
+	ConfigResolver Resolver = "config"
```
prateek: nit: only the first var in the const block needs the type specifier

config/hostid/hostid.go
```diff
@@ -0,0 +1,81 @@
...
+//
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package hostid
```
prateek: could you add a package comment to stay consistent with what Jerome's been adding to m3x

robskillington: Sure thing, done.

PR #108, opened 2017-10-22
Remove field slice in nulllogger to make it comparable

xichen2020:
cc @cw9 @robskillington @prateek @jeromefroe 

This PR makes null logger a pointer type.

### DIFF
--- a/log/logger.go
+++ b/log/logger.go
@@ -123,40 +123,23 @@ func (f Fields) Len() int { return len(f) }
 func (f Fields) ValueAt(i int) Field { return f[i] }
 
 // NullLogger is a logger that emits nowhere.
-var NullLogger Logger = &nullLogger{}
+var NullLogger Logger = nullLogger{}
 
-type nullLogger struct {
-	fields Fields
-}
+type nullLogger struct{}
 
-func (*nullLogger) Enabled(_ Level) bool                   { return false }
-func (*nullLogger) Fatalf(msg string, args ...interface{}) {}
-func (*nullLogger) Fatal(msg string)                       { os.Exit(1) }
-func (*nullLogger) Errorf(msg string, args ...interface{}) {}
-func (*nullLogger) Error(msg string)                       {}
-func (*nullLogger) Warnf(msg string, args ...interface{})  {}
-func (*nullLogger) Warn(msg string)                        {}
-func (*nullLogger) Infof(msg string, args ...interface{})  {}
-func (*nullLogger) Info(msg string)                        {}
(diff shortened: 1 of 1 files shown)


log/logger.go
```diff
@@ -155,7 +155,7 @@ func (l nullLogger) WithFields(newFields ...Field) Logger {
 		fields = append(fields, existingFields.ValueAt(i))
 	}
 	fields = append(fields, newFields...)
-	return nullLogger{Fields(fields)}
+	return &nullLogger{Fields(fields)}
```
jeromefroe: If we return a new logger here, don't we lose the ability to compare `NullLogger`s?

xichen2020: Yeah good point, in that case I'll remove the fields from the `NullLogger` struct since it's not logging anything regardless.

log/logger.go
```diff
@@ -123,26 +123,26 @@ func (f Fields) Len() int { return len(f) }
...
-func (nullLogger) Warn(msg string)                        {}
-func (nullLogger) Infof(msg string, args ...interface{})  {}
-func (nullLogger) Info(msg string)                        {}
-func (nullLogger) Debugf(msg string, args ...interface{}) {}
-func (nullLogger) Debug(msg string)                       {}
-func (l nullLogger) Fields() LoggerFields                 { return l.fields }
-
-func (l nullLogger) WithFields(newFields ...Field) Logger {
+func (*nullLogger) Enabled(_ Level) bool                   { return false }
+func (*nullLogger) Fatalf(msg string, args ...interface{}) {}
```
prateek: why not os.Exit(1) here too?

log/logger.go
```diff
@@ -123,26 +123,26 @@ func (f Fields) Len() int { return len(f) }
...
+func (*nullLogger) Fatal(msg string)                       { os.Exit(1) }
```
prateek: maybe log the message before exiting

log/logger.go
xichen2020: Not sure, that wasn't my change, I can add it however.

log/logger.go
xichen2020: Not sure if that's consistent with what the null logger does, IMO null logger shouldn't log anything.

log/logger.go
prateek: Sounds good

log/logger.go
prateek: Just feels weird to be crashing the process without any indication why.

xichen2020: I'd think in that case the user should not be using the null logger and instead use a normal logger like `SimpleLogger`.

PR #111, opened 2017-11-08
Add time.Time matcher and fix broken comparison

richardartoul:
1) Add a time.Time matcher for use with gomock
2) Fix IsEmpty method of range to use .Equal() method instead of ==


time/time_matcher.go
```diff
@@ -0,0 +1,34 @@
+package time
```
xichen2020: nit: just call it `matcher.go` since it's in the time package?

PR #114, opened 2017-11-14
[instrument] Add metric for GC CPU Fraction

jeromefroe:
Add a metric for `GCCPUFraction`, which tracks the percentage of total available CPU time that was spent on garbage collection.

cc @xichen2020 @prateek


prateek:
looks like theres an issue with the new linter config?

jeromefroe:
Yea, I think I'll sync with Ritchie tomorrow.

PR #115, opened 2017-11-22
Add resource pool for long lives objects

nikunjgit:


robskillington:
We already have one of these in `pool`?

PR #116, opened 2017-12-11
Add ability to specify per bucket options for a bucketized pool

robskillington:


pool/types.go
```diff
@@ -107,6 +107,11 @@ type Bucket struct {
 
 	// Count is the number of fixed elements in the bucket.
 	Count int
+
+	// Options is an optional override to specify options to use for a bucket,
+	// specify nil to use the options specified to the bucketized pool
+	// constructor for this bucket.
+	Options ObjectPoolOptions
```
prateek: suggestion(take/leave): `OverrideOptions` would make the use of this clearer

robskillington: So much cleaner though as just "Options" ha.  Yeah... I'll leave for now, it's a reasonable suggestion though.

PR #120, opened 2018-01-31
Add `context`, `ident` and `resource` package

prateek:
- Move `Finalizer` interface to newly created `resource` package (from `checked`)
- Migrate `m3db/context` -> `m3x/context`
- Merge `m3db/ts` and `m3x/id` as `m3x/ident`

/cc @robskillington


CONTRIBUTING.md
```diff
@@ -16,7 +16,7 @@ make test
 ## Making A Change
 
 *Before making any significant changes, please [open an
-issue](https://github.com/m3db/m3db/issues).* Discussing your proposed
+issue](https://github.com/m3db/m3x/issues).* Discussing your proposed
```
robskillington: Ha, good catch 👍

ident/types.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// A Datapoint is a single data value reported at a given time.
+type Datapoint struct {
+	Timestamp time.Time
+	Value     float64
+}
+
+// Annotation represents information used to annotate datapoints.
+type Annotation []byte
```
robskillington: Oh Datapoint and Annotation should probably stay in `m3db` - they truly do belong in a `ts` (timeseries) package and not in an `ident` package.

robskillington:
We need to work out how to generate `ident_mock.go` from m3x now too, copying the mock in won't cut it I don't think... unless we truly don't expect to add other types we want to mock, etc later.

prateek:
opened #122 as a followup

PR #121, opened 2018-02-01
Add Tag and IDsIterator to `ident`

prateek:
/cc @robskillington


ident/types.go
```diff
@@ -43,24 +43,69 @@ type ID interface {
...
+type TagValue ID
+
+// Tag represents a timeseries tag.
+type Tag struct {
+	Name  TagName
+	Value TagValue
+}
+
 // IdentifierPool represents an automatic pool of IDs.
 type IdentifierPool interface {
```
prateek: How do you feel about renaming this to `ident.Pool`?

robskillington: Sure thing.

prateek: Awesome, done

PR #123, opened 2018-02-03
Add `ident` utility methods

prateek:
/cc @robskillington


prateek:
Yep it's to allow multiple iteration. Needed the ability to either Clone() or Reset() the head of the iterator for this [code](https://github.com/m3db/m3db/blob/262755e40fdc6dd08bbbdc6850b8f0bb9b7b0b0f/client/session.go#L863-L885). Chose Clone() cause it felt more in line with the iterator idiom

PR #124, opened 2018-02-16
Add Equal() method to Tags

prateek:


ident/tags_test.go
```diff
@@ -0,0 +1,68 @@
...
+	tagsB := Tags{
+		StringTag("foo", "bar"),
+		StringTag("and", "done"),
+	}
+
+	require.False(t, tagsA.Equal(tagsB))
+	require.False(t, tagsB.Equal(tagsA))
+}
+
+func TestTagsInequalOrder(t *testing.T) {
```
jeromefroe: nit: s/TestTagsInequalOrder/TestTagsUnequalOrder

ident/types.go
```diff
@@ -143,6 +143,28 @@ func (ids IDs) Finalize() {
...
 
+// Finalize finalizes all Tags.
+func (tags Tags) Finalize() {
+	for _, t := range tags {
+		t.Name.Finalize()
+		t.Value.Finalize()
+	}
+}
+
+// Equal returns a bool indicating if the tags are equal.
```
jeromefroe: nit: perhaps it's worth calling out that the tags must be in the same order?

ident/tags_test.go
prateek: done

ident/types.go
prateek: sure thing

PR #125, opened 2018-02-20
Add utility methods to ident.Pool

prateek:
/cc @robskillington

### DIFF
--- a/ident/iterator.go
+++ b/ident/iterator.go
@@ -71,7 +71,7 @@ func (i *idSliceIter) Remaining() int {
 	return 0
 }
 
-func (i *idSliceIter) Clone() Iterator {
+func (i *idSliceIter) Duplicate() Iterator {
 	return &idSliceIter{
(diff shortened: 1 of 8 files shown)


ident/iterator.go
```diff
@@ -86,29 +90,60 @@ func NewStringIDsIterator(ids ...string) Iterator {
...
 	}
-	i.currentID = nil
+	i.release()
 	return false
 }
 
+func (i *stringSliceIter) setCurrent(s string) {
+	// we already hold a reference to the underlying bytes over the lifecycle
+	// of the iterator, don't need another one here.
+	i.idBytes.Reset([]byte(s))
```
robskillington: Hm (as discussed), maybe too much work to avoid extra allocs that will still ultimately require a byte slice alloc here.

prateek: Yep that's fair. Reverted in the next version.

PR #127, opened 2018-02-26
[glide] Update glide.yaml to use semver where possible

jeromefroe:
I ran into an issue in `m3ninx` that required me to update to a newer version of the `github.com/golang/mock` package but was unable to because `m3x` pins it to an older SHA. Consequently, this diff updates our dependency on `mock` in Glide to use semver releases. I also noticed there were a few other packages where we could use semver instead so I took the liberty to update those as well so we can avoid this situation in the future as well.

cc @prateek


glide.yaml
```diff
@@ -29,22 +23,31 @@ import:
...
+  version: ^1.2.0
 
 - package: github.com/shirou/w32
   version: bb4de0191aa41b5507caa14b0650cdbddcd9280b
 
 - package: github.com/StackExchange/wmi
-  version: e542ed97d15e640bdc14b5c12162d59e8fc67324
+  version: ^1.0.0
 
 - package: github.com/fortytw2/leaktest
```
prateek: can this be moved to testImport too?

PR #130, opened 2018-03-09
Add age reporting in instrument.BuildReporter

prateek:
/cc @robskillington @jeromefroe

### DIFF
--- a/instrument/build.go
+++ b/instrument/build.go
@@ -25,6 +25,7 @@ import (
 	"log"
 	"runtime"
 	"sync"
+	"sync/atomic"
 	"time"
 )
 
@@ -41,8 +42,8 @@ var (
 	// BuildDate is the date this build was created.
 	BuildDate = "unknown"
 
-	// BuildDateUnixNanos is the UnixNanos since epoch representing the date this build was created.
-	BuildDateUnixNanos int64 = 0
+	// BuildTimeUnixNanos is the UnixNanos since epoch representing the date this build was created.
+	BuildTimeUnixNanos int64
 
 	// LogBuildInfoAtStartup controls whether we log build information at startup. If its
 	// set to a non-empty string, we log the build information at process startup.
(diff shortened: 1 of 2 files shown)


instrument/build.go
```diff
@@ -59,10 +65,11 @@ var (
...
 func LogBuildInfo() {
-	log.Printf("Go Runtime version: %s\n", goVersion)
-	log.Printf("Build Revision:     %s\n", Revision)
-	log.Printf("Build Branch:       %s\n", Branch)
-	log.Printf("Build Date:         %s\n", BuildDate)
+	log.Printf("Go Runtime version:  %s\n", goVersion)
+	log.Printf("Build Revision:      %s\n", Revision)
+	log.Printf("Build Branch:        %s\n", Branch)
+	log.Printf("Build Date:          %s\n", BuildDate)
+	log.Printf("Build DateUnixNanos: %s\n", BuildDateUnixNanos)
```
xichen2020: nit: use `%d` instead of `%s` for int64s.

instrument/build.go
```diff
@@ -41,15 +41,21 @@ var (
 	// BuildDate is the date this build was created.
 	BuildDate = "unknown"
 
+	// BuildDateUnixNanos is the UnixNanos since epoch representing the date this build was created.
+	BuildDateUnixNanos int64 = 0
```
xichen2020: nit: `BuildTimeUnixNanos` might be more apt. Also this should probably be a const?

instrument/build.go
prateek: sure thing

instrument/build.go
prateek: sure thing, will rename.

Re `const`: nah, needs to be a `var` to allow it to be overriden from ldflags at build time.

xichen2020: Ah yeah that's fair, might need to get rid of the 0 initial to silence lint.

instrument/build.go
prateek: fixed

instrument/build.go
prateek: yep sure thing. fixed both issues.

PR #131, opened 2018-03-10
Fix age reporter

prateek:
Turns out LDFlags need to be strings. Porting the unix nanos number to a string

### DIFF
--- a/instrument/build.go
+++ b/instrument/build.go
@@ -100,7 +100,7 @@ func NewBuildReporter(
 }
 
 func (b *buildReporter) Start() error {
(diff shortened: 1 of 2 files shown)


instrument/build.go
```diff
@@ -98,11 +100,25 @@ func NewBuildReporter(
 }
 
 func (b *buildReporter) Start() error {
+	var (
```
xichen2020: nit: should probably be `const`.

prateek: sure thing, addressed in next patch.

PR #132, opened 2018-03-19
Remove unused and problematic dependencies (license, missing glide, etc)

robskillington:


glide.lock
```diff
@@ -55,22 +36,19 @@ imports:
...
-  version: f6cff0780e542efa0c8e864dc8fa522808f6a598
-  subpackages:
-  - unix
-  - windows
 - name: gopkg.in/validator.v2
   version: 3e4f037f12a1221a0864cf0dd2e81c452ab22448
 - name: gopkg.in/yaml.v2
   version: a83829b6f1293c91addabc89d0571c246397bbf4
 testImports:
 - name: github.com/davecgh/go-spew
```
prateek: looks like you deleted it in the glide.yaml, guessing its needed by some transitive package?

robskillington: Yeah =[

PR #133, opened 2018-03-22
Add generic value type map for ident.ID and []byte key based maps

robskillington:
This adds a generic value type map that we can use for anything package using the `ident.ID` type.

You can generate the source for the map using:
```sh
VALUE_TYPE=AnyType go generate
```

cc @jeromefroe @prateek @xichen2020


ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+// ValueType is the generic value type for use with the specialized hash map.
+type ValueType generic.Type
+
+// Map uses the genny package to provide a generic map that can be specialized
+// by running the following command from this directory:
+// ```
+// VALUE_TYPE=YourValueType go generate
+// ```
+// This will output to stdout the generated source file to use for your map.
+// It uses linear probing by incrementing the number of the hash created when
```
prateek: Would be good to include benchmark for expected performance.

ident/map_gen.go
```diff
@@ -0,0 +1,26 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package ident
+
+// This generate command will generate a new specialized ident keyed map,
+// you can use it with:
+// VALUE_TYPE=YourValueType go generate
+//go:generate sh -c "set -x; (which genny >/dev/null || (go get -u github.com/cheekybits/genny && go install github.com/cheekybits/genny)) && cat ./map.go | genny gen \"ValueType=${VALUE_TYPE}\""
```
prateek: might be better to package this is a script rather than having people copy paste this command

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+}
+
+// Len returns the number of map entries in the map.
+func (m *Map) Len() int {
+	return len(m.lookup)
+}
+
+// Has returns true if value exists for key, false otherwise, it is
+// shorthand for a call to Get that doesn't return the value.
+func (m *Map) Has(id ID) bool {
```
prateek: naming nit: `Contains` instead of `Has`

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+	}
+
+	m.lookup[hash] = MapEntry{
+		key:   key,
+		value: value,
+	}
+}
+
+// Iter provides the underlying map to allow for using a native Go for loop
+// to iterate the map, however callers should only ever read and not write
```
prateek: super nit: s/write/modify

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+func (m *Map) ResetByReallocateMap() {
+	if m.opts.Size > 0 {
+		m.lookup = make(map[MapHash]MapEntry, m.opts.Size)
+	} else {
+		m.lookup = make(map[MapHash]MapEntry)
+	}
+}
+
+// inline alloc-free fnv-1a hash
+func hashFnv(key ID) MapHash {
```
prateek: maybe add a few tests to ensure this gives the same results as the implementation in `hash/fnv`

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+// VALUE_TYPE=YourValueType go generate
+// ```
+// This will output to stdout the generated source file to use for your map.
+// It uses linear probing by incrementing the number of the hash created when
+// hashing the identifier if there is a collision. The hashing algorithm used is
+// fnv hash.
+// The map is a value type and not an interface to allow for less painful
+// upgrades when adding/removing methods, it is not likely to need mocking so
+// an interface would not be super useful either.
+type Map struct {
```
prateek: Do you envision other implementations added in as and when they're needed (e.g. your list + map version to enable stable iteration over function boundaries).

ident/map.go
prateek: Orthogonal question: would it be worth it to benchmark different approaches (chaining/double-hashing)?

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+
+	"github.com/cheekybits/genny/generic"
+)
+
+const (
+	offset64 = 14695981039346656037
+	prime64  = 1099511628211
+)
+
+// ValueType is the generic value type for use with the specialized hash map.
```
prateek: How about also accepting a generic Key type, and functions to resolve to appropriate values in the ctor.  Something like:

```
type KeyType generic.Type
type HashFn(KeyType) MapHash
type CopyFn(KeyType) KeyType
type EqualsFn(KeyType, KeyType) bool

type MapOptions struct {
  Size int64 
  HashFn HashFn
  CopyFn CopyFn
  EqualsFn EqualsFn
} 
```

Could extract that map implementation into its `m3x/generic`, and any users would specialize both key/value types

prateek: PS this is beginning to look [familiar](http://www.cplusplus.com/reference/map/map/)

```
template < class Key,                                     // map::key_type
           class T,                                       // map::mapped_type
           class Compare = less<Key>,                     // map::key_compare
           class Alloc = allocator<pair<const Key,T> >    // map::allocator_type
           > class map;
```

ident/map.go
prateek: maybe rename to `HashMap` in case we decide to add an ordered map implementation down the road.

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+		}
+		// Linear probe to "next" to this entry (really a rehash)
+		hash++
+	}
+	var empty ValueType
+	return empty, false
+}
+
+// Set will set the value for an identifier.
+func (m *Map) Set(id ID, value ValueType) {
```
xichen2020: Would it be cleaner to consolidate `Set` and `SetNoCopyNoFinalize` into one API, e.g., something like
```
type SetOptions struct {
  CopyID bool
}

func (m *Map) Set(id ID, value ValueType, o SetOptions) {
  var opts mapSetOptions
  if o.CopyID {
    opts = mapSetOptions{
      copyKey: true,   
      finalizeKey: true,
    })
  }
  m.set(id, opts)
} 
```
?

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+	m.lookup[hash] = MapEntry{
+		key:   key,
+		value: value,
+	}
+}
+
+// Iter provides the underlying map to allow for using a native Go for loop
+// to iterate the map, however callers should only ever read and not write
+// the map.
+func (m *Map) Iter() map[MapHash]MapEntry {
```
xichen2020: Would it be better to call this something like `Iterable`? Feels like a function named `Iter` should return an iterator instead.
```
For k, v := range m.Iterable()
```
seems more intuitive than
```
For k, v := range m.Iter()
```

ident/map.go
xichen2020: +1

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+func (m *Map) Get(id ID) (ValueType, bool) {
+	hash := m.hash(id)
+	for v, ok := m.lookup[hash]; ok; v, ok = m.lookup[hash] {
+		if v.key.id.Equal(id) {
+			return v.value, true
+		}
+		// Linear probe to "next" to this entry (really a rehash)
+		hash++
+	}
+	var empty ValueType
```
xichen2020: nit: simply `return ValueType{}, false`? Compiler should do a decent job at return value optimization.

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+		if v.key.finalize {
+			v.key.id.Finalize()
+		}
+	}
+}
+
+// ResetByReallocateMap will avoid deleting all keys and reallocate
+// a new map, this is useful if you believe you have a larage map and
+// will not need to grow back to a similar size.
+func (m *Map) ResetByReallocateMap() {
```
xichen2020: Is this really needed? Feels like users could just call the ctor if they need to call this, and this is probably not going to be called that often so the overhead of one extra pointer is probably a worthy tradeoff for smaller API surface area.

ident/map.go
```diff
@@ -0,0 +1,256 @@
...
+
+package ident
+
+import (
+	"github.com/m3db/m3x/checked"
+
+	"github.com/cheekybits/genny/generic"
+)
+
+const (
```
xichen2020: nit: Might be cleaner to move the const defs into the `fnvHash` method since they are only useful there and nowhere else.

ident/map.go
robskillington: The thing I didn't like about that was needing to pass in SetOptions even when you don't want to specify anything special.

ident/map.go
robskillington: So thing is, ValueType could be a native type like string which would code-gen `return string{}, false` which isn't valid.  Hence using the empty value declaration here.

ident/map.go
robskillington: They could call the constructor again yes, however I did want it to make it clear to the caller on what they were doing when they did Reset and give them an alternative that's very simple to avoid deleting all the keys if they really wanted to.

generics/hashmap/byteskey/new_map.go
```diff
@@ -0,0 +1,73 @@
...
+		}
+		finalizeFn = func(k []byte) {
+			pool.Put(k)
+		}
+	}
+	return newMap(mapOptions{
+		HashFn: func(k []byte) MapHash {
+			return MapHash(fnv.Hash64a(k))
+		},
+		EqualsFn: func(x, y []byte) bool {
```
prateek: nit: could simplify to `EqualsFn: bytes.Equal`

generics/hashmap/idkey/map.go
```diff
@@ -0,0 +1,249 @@
...
+// make hashmap key_type=YouridentID value_type=YourValue
+// ```
+// This will output to stdout the generated source file to use for your map.
+// It uses linear probing by incrementing the number of the hash created when
+// hashing the identifier if there is a collision.
+// Map is a value type and not an interface to allow for less painful
+// upgrades when adding/removing methods, it is not likely to need mocking so
+// an interface would not be super useful either.
+type Map struct {
+	mapOptions
```
prateek: same comments as the byteskey map

generics/hashmap/idkey/new_map.go
```diff
@@ -0,0 +1,108 @@
...
+		CopyFn:     copyFn,
+		FinalizeFn: finalizeFn,
+		Size:       opts.Size,
+	})
+}
+
+// id is a small utility type to avoid the heavy-ness of the true ident.ID
+// implementation when copying keys just for the use of looking up keys in
+// the map internally.
+type id []byte
```
prateek: maybe add a var decl to ensure `id` satisfies the `ident.ID` interface.

generics/hashmap/idkey/new_map.go
```diff
@@ -0,0 +1,108 @@
...
+
+func (v id) Data() checked.Bytes {
+	panic("not implemented")
```
prateek: Could you add a comment explaining this is deliberate.

generics/hashmap/map.go
```diff
@@ -0,0 +1,257 @@
...
+// make idhashmap pkg=outpackage value_type=Type out_dir=/tmp
+// ```
+// This will output to stdout the generated source file to use for your map.
+// It uses linear probing by incrementing the number of the hash created when
+// hashing the identifier if there is a collision.
+// Map is a value type and not an interface to allow for less painful
+// upgrades when adding/removing methods, it is not likely to need mocking so
+// an interface would not be super useful either.
+type Map struct {
+	mapOptions
```
prateek: same comment as the other gen'd maps

ident/identifier.go
```diff
@@ -43,68 +41,32 @@ func StringID(str string) ID {
 
 type id struct {
 	data checked.Bytes
-	hash Hash
```
prateek: Any concern about removing this, or is fnv64 fast enough that it doesn't matter if we compute hashes multiple times?

generics/hashmap/byteskey/map.go
```diff
@@ -0,0 +1,247 @@
...
+// make hashmap key_type=Your[]byte value_type=YourValue
+// ```
+// This will output to stdout the generated source file to use for your map.
+// It uses linear probing by incrementing the number of the hash created when
+// hashing the identifier if there is a collision.
+// Map is a value type and not an interface to allow for less painful
+// upgrades when adding/removing methods, it is not likely to need mocking so
+// an interface would not be super useful either.
+type Map struct {
+	mapOptions
```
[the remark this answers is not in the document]
robskillington: It's not a ref to the original struct, it's copy semantics so they won't be able to change the internals after they've passed it in.

generics/hashmap/byteskey/new_map.go
robskillington: Sure thing.

generics/hashmap/byteskey/map.go
```diff
@@ -0,0 +1,247 @@
...
+	// EqualsFn is the equals key function to execute when detecting equality.
+	EqualsFn EqualsFn
+	// CopyFn is the copy key function to execute when copying the key.
+	CopyFn CopyFn
+	// FinalizeFn is the finalize key function to execute when finished with a
+	// key, this is optional to specify.
+	FinalizeFn FinalizeFn
+	// Size is the initial size for the map, use zero to use Go map initial size
+	// and consequently is optional to specify.
+	Size int
```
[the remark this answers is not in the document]
robskillington: Sure thing.

generics/hashmap/byteskey/gen.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+
+set -x
+set -e
+
+PKG="github.com/cheekybits/genny"
+
+which genny >/dev/null || (go get -u $PKG && go install $PKG)
```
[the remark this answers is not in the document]
robskillington: Not necessarily, maybe I'll move these into the makefile (this is to update the byteskey and idkey packages themselves, not the code they generate).

Makefile
```diff
@@ -89,6 +89,24 @@ mock-gen: install-mockgen install-license-bin install-util-mockclean
 	@echo Generating mocks
 	PACKAGE=$(m3x_package) $(auto_gen) $(mocks_output_dir) $(mocks_rules_dir)
 
+.PHONY: install-genny
+install-genny:
+	@which genny >/dev/null || (go get -u github.com/cheekybits/genny && go install github.com/cheekybits/genny)
```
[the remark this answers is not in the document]
robskillington: Sure, can do.

generics/hashmap/idkey/map.go
robskillington: It's not a ref to the original struct, it's copy semantics so they won't be able to change the internals after they've passed it in.

generics/hashmap/idkey/new_map.go
robskillington: Good idea, will do.

generics/hashmap/idkey/new_map.go
robskillington: Sure thing.

generics/hashmap/map.go
robskillington: It's not a ref to the original struct, it's copy semantics so they won't be able to change the internals after they've passed it in.

generics/hashmap/byteskey/map.go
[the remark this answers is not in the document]
prateek: I thought this would still break that:

```
m := NewMap(...)  // returns *Map
m.HashFn = ...
```

ident/identifier.go
prateek: Fair enough, that seems reasonable to me. 

As an aside: why fnv over xx? All the stuff I can find online suggests xxhash is all around better for our kinda workload (e.g. [1](https://aras-p.info/blog/2016/08/02/Hash-Functions-all-the-way-down/) [2](https://github.com/Cyan4973/xxHash/)).

robskillington: Hm, yeah xxhash does seem to be quite fast:
```go
$ go test -v -bench . -run none
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3x/hash/fnv
BenchmarkFnv-4          30000000                54.3 ns/op
BenchmarkMd5-4          10000000               165 ns/op
BenchmarkXxhash-4       100000000               20.3 ns/op
PASS
ok      github.com/m3db/m3x/hash/fnv    5.583s
```

Also seems to be used in Influx and Prometheus:
https://github.com/cespare/xxhash#projects-using-this-package

I'll pull it in, MIT licensed so should be good.

generics/hashmap/byteskey/map.go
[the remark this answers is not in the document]
robskillington: Yeah it does seem to export these, strange thought it wouldn't considering the enclosing struct type is not exported.

PR #134, opened 2018-03-26
Improve object pool performance

cw9:
On a frequently accessed pool, we actually spend a lot of time on calling time.Now(), this change should save 600-700ns per operation.

@xichen2020 @robskillington @prateek @jeromefroe

### DIFF
--- a/pool/object.go
+++ b/pool/object.go
@@ -24,7 +24,6 @@ import (
 	"errors"
 	"math"
 	"sync/atomic"
-	"time"
 
 	"github.com/uber-go/tally"
 )
@@ -49,6 +48,7 @@ type objectPool struct {
 	refillHighWatermark int
 	filling             int32
 	initialized         int32
+	dice                int32
 	metrics             objectPoolMetrics
 }
 
@@ -145,7 +145,7 @@ func (p *objectPool) Put(obj interface{}) {
 }
 
 func (p *objectPool) trySetGauges() {
-	if time.Now().UnixNano()%sampleObjectPoolLengthEvery == 0 {
+	if atomic.AddInt32(&p.dice, 1)%sampleObjectPoolLengthEvery == 0 {
 		p.setGauges()
 	}
 }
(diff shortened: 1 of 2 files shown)


pool/object.go
```diff
@@ -145,7 +145,7 @@ func (p *objectPool) Put(obj interface{}) {
 }
 
 func (p *objectPool) trySetGauges() {
-	if time.Now().UnixNano()%sampleObjectPoolLengthEvery == 0 {
+	if atomic.AddInt32(&p.dice, 1)%sampleObjectPoolLengthEvery == 0 {
```
prateek: mind adding a quick benchmark repro-ing the issue - will make it easier to test future changes.

cw9: Sounds good

robskillington:
@cw9 are you sure you were fairly testing this and not running into this bug?
https://github.com/golang/go/issues/22601

robskillington:
After upgrading to Go 1.10 to avoid that bug I see it's just slightly faster but not that much, but will require flushing the CPU cache line because we'll be doing an atomic increment instead of just a read.
```
r@pool[master]$ go test -v -benchmem -run none -bench BenchmarkObjectPoolGetPut
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3x/pool
BenchmarkObjectPoolGetPut-4     20000000               100.0 ns/op             0 B/op          0 allocs/op
PASS
ok      github.com/m3db/m3x/pool        2.131s
r@pool[master]$
r@pool[master]$ go test -v -benchmem -run none -bench BenchmarkObjectPoolGetPut
goos: darwin
goarch: amd64
pkg: github.com/m3db/m3x/pool
BenchmarkObjectPoolGetPut-4     10000000               132 ns/op               0 B/op          0 allocs/op
PASS
ok      github.com/m3db/m3x/pool        1.498s
```

It might be worth running this in prod on something that utilizes pooling very heavily?

PR #140, opened 2018-04-20
Add ident.TagIteratorMatcher

prateek:


ident/tag_iterator_matcher.go
```diff
@@ -0,0 +1,78 @@
...
+
+// TagIteratorMatcher is a gomock.Matcher that matches TagIterator
+type TagIteratorMatcher interface {
+	gomock.Matcher
+}
+
+// NewTagIteratorMatcher returns a new TagIteratorMatcher
+func NewTagIteratorMatcher(inputs ...string) TagIteratorMatcher {
+	if len(inputs)%2 != 0 {
+		panic("inputs must be specified in name-value pairs (i.e. divisible by 2)")
```
robskillington: nit: Perhaps just return `(TagIteratorMatcher, error)` instead of panic (I know this is used in tests, but even then a `require.NoError(t, err)` at least doesn't stop executing the rest of the package tests and fails cleanly.

xichen2020: Btw a common pattern to deal w/ this is to have both `NewTagIteratorMatcher` that returns an error and a `MustNewTagIteratorMatcher` that panics on invalid input.

cw9: Or maybe rename to `MustNew...` and panic

ident/tag_iterator_matcher.go
```diff
@@ -0,0 +1,78 @@
...
+		return false
+	}
+	if t.Err() != nil {
+		return false
+	}
+	return true
+}
+
+func (m *tagIterMatcher) String() string {
+	return fmt.Sprintf("tagIter %+v", m.inputs)
```
xichen2020: nit: probably `%v` is sufficient.

ident/tag_iterator_matcher.go
prateek: Sure i'll do the two method approach.

ident/tag_iterator_matcher.go
prateek: fair

prateek: Fixed in next version.

ident/tag_iterator_matcher.go
prateek: Done

PR #141, opened 2018-04-22
 Create generic arraypool

prateek:
- Adds new Genny target for an ArrayPool 
- Related makefile scripts and so on


generics/arraypool/doc.go
```diff
@@ -0,0 +1,31 @@
...
+
+import (
+	"github.com/cheekybits/genny/generic"
+)
+
+// elemType is the generic type for use with the specialized array pool.
+// NB: we need this type decl to allow codegen and tests to succeed. It's
+// kept in a separate file to allow generation of specialized array pools
+// without requiring a corresponding element declaration.
+type elemType generic.Type
```
xichen2020: Hm not sure if I follow, what happens if you put this typedef in `pool.go`?

context/finalizers_arraypool_gen.go
```diff
@@ -0,0 +1,102 @@
...
+	})
+}
+
+func (p *finalizersArrPool) Get() []resource.Finalizer {
+	return p.pool.Get().([]resource.Finalizer)
+}
+
+func (p *finalizersArrPool) Put(arr []resource.Finalizer) {
+	arr = p.opts.FinalizeFn(arr)
+	if max := p.opts.MaxCapacity; max > 0 && cap(arr) > max {
```
xichen2020: nit: I'd just remove `max > 0` here.

generics/arraypool/pool.go
```diff
@@ -0,0 +1,100 @@
...
+	var empty elemType
+	for i := range elems {
+		elems[i] = empty
+	}
+	elems = elems[:0]
+	return elems
+}
+
+// nolint
+type elemArr []elemType
```
xichen2020: Is `elemArr` and `grow` used anywhere? Can't seem to find where they are used other than in the tests. If so, I'd suggest moving them to the test files to be clear of the scope.

generics/arraypool/doc.go
```diff
@@ -0,0 +1,31 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package arraypool
+
+import (
+	"github.com/cheekybits/genny/generic"
```
xichen2020: fwiw I've been using `https://github.com/mauricelam/genny` instead, which supports a number of features that the original repo doesn't support, in particular
* User-defined types can be specified for generic types.
* You can specify that generic type should implement some interfaces.

The original repo doesn't seem to be actively maintained (last update was Mar 2017)

Makefile
```diff
@@ -126,37 +126,75 @@ endif
...
-		gorename -from '\"$(codegen_package)\".CopyFn' -to $(rename_type_prefix)MapCopyFn && \
-		gorename -from '\"$(codegen_package)\".FinalizeFn' -to $(rename_type_prefix)MapFinalizeFn && \
-		gorename -from '\"$(codegen_package)\".MapEntry' -to $(rename_type_prefix)MapEntry && \
-		gorename -from '\"$(codegen_package)\".SetUnsafeOptions' -to $(rename_type_prefix)MapSetUnsafeOptions && \
-		gorename -from '\"$(codegen_package)\".mapAlloc' -to _$(rename_type_prefix)MapAlloc && \
-		gorename -from '\"$(codegen_package)\".mapOptions' -to _$(rename_type_prefix)MapOptions && \
-		gorename -from '\"$(codegen_package)\".mapKey' -to _$(rename_type_prefix)MapKey && \
-		gorename -from '\"$(codegen_package)\".mapKeyOptions' -to _$(rename_type_prefix)MapKeyOptions && \
-		bash -c 'test \"$(rename_constructor)\" = \"\" || gorename -from \"\\\"$(codegen_package)\\\".NewMap\" -to $(rename_constructor)' && \
+	docker run --rm -it                                                                                                                                        \
```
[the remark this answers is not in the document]
prateek: Yep, they are. Here's an easier view to tell: https://github.com/m3db/m3x/blob/3bdebd2ed3669c849e5295ce0ec4bf7fc051f1a2/Makefile 

PS https://github.com/godlygeek/tabular is amazing for this kinda work if you're a vim user.

context/finalizers_arraypool_gen.go
prateek: Hm i was using 0 as a sentinel to allow user to disable max capacity checking. You reckon it's not worth doing so?

generics/arraypool/doc.go
prateek: Nice! especially (2) is super useful. Will move to it in the next patch

generics/arraypool/pool.go
prateek: Yep not used in the current diff directly, included here because it's related to the slice type. E.g. https://github.com/m3db/m3db/blob/64f3241fed93fffecf484c73f43a8b478982bd9e/network/server/tchannelthrift/node/service.go#L1069 

It's better to have a single place for all the testing in case we need to make changes to the behaviour down the road. I'd prefer to leave here unless you feel strongly.

generics/arraypool/doc.go
prateek: Didn't work with other genny but seems to be fine with the one you suggested we use. Will remove it.

generics/arraypool/pool_test.go
```diff
@@ -0,0 +1,92 @@
...
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package arraypool
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
```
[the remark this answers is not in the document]
prateek: fixed

generics/arraypool/doc.go
prateek: done

generics/arraypool/doc.go
prateek: done

Makefile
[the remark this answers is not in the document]
xichen2020: noice =)

context/finalizers_arraypool_gen.go
xichen2020: Yeah was thinking users could set capacity to `math.MaxInt64` if they want to disable max capacity checking to avoid using `0` as a special value, but I think we might already be doing in other places, so I'm fine w/ leaving it as is for consistency's sake.

generics/arraypool/pool.go
xichen2020: I see, makes sense.

Do you envision the pool/array type will only be used as internal types/methods? It means the generated types will have to be placed in the same package where they are used and not outside the package. Not sure if that's too strong a restriction in terms of repo layout etc.

context/finalizers_arraypool_gen.go
prateek: Sounds good

generics/arraypool/pool.go
prateek: I think it'll be internal only for majority of the cases but people will occasionally want to have an exported version. We support that through renaming (e.g. [steps](https://github.com/m3db/m3x/blob/fab025948308e135c2d03d1369e0ccca2e2f6e1e/Makefile#L172-L173)). 

TBH: I haven't tested creating an exported version yet, so the first person who does it will have to a make change or two for more renaming I'm sure.

PR #143, opened 2018-04-24
Update genny generation to use forked gorename

prateek:
- no more docker as a ci build dep

### DIFF
--- a/generated-source-files.mk
+++ b/generated-source-files.mk
@@ -1,0 +1,139 @@
+gorename_package         := github.com/prateek/gorename
+gorename_package_version := 52c7307cddd221bb98f0a3215216789f3c821b10
+
+# Tests that all currently generated types match their contents if they were regenerated
+.PHONY: test-genny-all
+test-genny-all: genny-all
+	@test "$(shell git diff --shortstat 2>/dev/null)" = "" || (git diff --no-color && echo "Check git status, there are dirty files" && exit 1)
+	@test "$(shell git status --porcelain 2>/dev/null | grep "^??")" = "" || (git status --porcelain && echo "Check git status, there are untracked files" && exit 1)
+
+.PHONY: genny-all
+genny-all: genny-map-all genny-arraypool-all
+
+.PHONY: genny-map-all
+genny-map-all: idhashmap-update byteshashmap-update
+
+.PHONY: idhashmap-update
+idhashmap-update: install-generics-bin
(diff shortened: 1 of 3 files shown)


generated-source-files.mk
```diff
@@ -0,0 +1,139 @@
...
+	$(eval out_dir=$(gopath_prefix)/$(target_package))
+	cd generics/hashmap/byteskey && cat ./map_gen.go | grep -v nolint | genny -pkg $(pkg) gen "MapValue=$(value_type)" > "$(out_dir:\=)/map_gen.go"
+	cd generics/hashmap/byteskey && cat ./new_map.go | grep -v nolint | genny -pkg $(pkg) gen "MapValue=$(value_type)" > "$(out_dir:\=)/new_map_gen.go"
+ifneq ($(rename_type_prefix),)
+	make hashmap-gen-rename
+endif
+
+.PHONY: hashmap-gen-rename-helper
+hashmap-gen-rename-helper:
+	gorename -from '"$(target_package)$(temp_suffix)".Map' -to $(rename_type_prefix)Map
```
robskillington: Can we declare a fixed `temp_suffix` so callers don't need to set it?  Also assume probably want to use `_tmp` instead of just `tmp` otherwise it'll be some strange target package like `../storage/repairtemp/...` rather than `../storage/repair_tmp/...`.

Can you also add a comment somewhere that `target_package` needs to avoid having a trailing slash?

xichen2020:
Just for context, why do we need to run it inside the container in the first place and why does the forked gorename fix the issue?

generated-source-files.mk
prateek: sure thing

prateek:
@xichen2020 upstream `gorename` by default scans all of `GOPATH` when either the  `-from` or `-to` targets are exported (known issue: https://github.com/golang/go/issues/10917). This makes running genny targets in-feasible locally where we have a massive number of repos in GOPATH typically. 

To get around this, we were running in docker (which had a much smaller GOPATH). That worked fine locally but the CI didn't have docker installed, so I forked `gorename` to never do the global scope checks it does for exported types. In theory this version of gorename isn't completely "safe" because it could cause conflicts, but in practice the package level checks it does are sufficient checks for use. And, it's like a million times faster.

generated-source-files.mk
prateek: done

PR #144, opened 2018-04-24
Calling init on worker pool twice no longer blocks

jskelcy:
While working on a service that used two worker pool I accidentally fat fingered and called init twice on the same worker pool. As a result, the service hung forever because we try to blindly push to an already full channel.

This diff checks to see if there are already workers in the channel and if there are just returns. I could see this going either way where we panic/error or silently return. If someone feels strongly about one way or the other I would happy to make a change.

### DIFF
--- a/sync/worker_pool.go
+++ b/sync/worker_pool.go
@@ -56,6 +56,9 @@ func NewWorkerPool(size int) WorkerPool {
 }
 
 func (p *workerPool) Init() {
+	if len(p.workCh) == cap(p.workCh) {
+		return
+	}
 	for i := 0; i < cap(p.workCh); i++ {
 		p.workCh <- struct{}{}
 	}
--- a/sync/worker_pool_test.go
+++ b/sync/worker_pool_test.go
@@ -121,3 +121,23 @@ func TestGoWithTimeout(t *testing.T) {
 
 	require.Equal(t, uint32(testWorkerPoolSize+1), count)
 }
+
+func TestFullWorkerPool(t *testing.T) {
+	var count uint32
+
+	p := NewWorkerPool(testWorkerPoolSize)
+	p.Init()
+	p.Init()
+
+	var wg sync.WaitGroup
+	for i := 0; i < testWorkerPoolSize*2; i++ {
+		wg.Add(1)
+		p.Go(func() {
+			atomic.AddUint32(&count, 1)
+			wg.Done()
+		})
+	}
+	wg.Wait()
+
+	require.Equal(t, uint32(testWorkerPoolSize*2), count)
+}


sync/worker_pool.go
```diff
@@ -56,6 +56,9 @@ func NewWorkerPool(size int) WorkerPool {
 }
 
 func (p *workerPool) Init() {
+	if len(p.workCh) == cap(p.workCh) {
```
jeromefroe: Since the length of the worker channel can change as the pool is used would it be better to instead use an internal flag and only initialize the channel if the flag is not?

jskelcy: The work channel is not public, so it would have to change within this package, and there is no code to do that right now. 

But in the future, I suppose it could change if some functionality was added. I can use a flag.

sync/worker_pool_test.go
```diff
@@ -121,3 +121,23 @@ func TestGoWithTimeout(t *testing.T) {
 
 	require.Equal(t, uint32(testWorkerPoolSize+1), count)
 }
+
+func TestFullWorkerPool(t *testing.T) {
```
cw9: Not exactly sure what is this testing, if you remove the second Init() does the test still pass? Also should probably rename this test to something like `TestInitOnInitializedWorkerPool`

sync/worker_pool.go
jeromefroe: I just meant that if one calls `Go` on the worker pool that will decrement the length of the channel until the worker is put back on the queue. So if there are any concurrent calls to `Go` between calls to `Init` you'll run into the same issue.

sync/worker_pool_test.go
jskelcy: yeah this test is a little odd because init does not return anything. I just wanted to show that this no longer hangs forever. Maybe the name change will make it clear.

robskillington:
Either that or we can just get rid of the init method altogether, it'll be more expensive to create the worker pool when using the constructor but everyone usually just calls Init straight afterwards anyhow.

sync/worker_pool_test.go
```diff
@@ -121,3 +121,23 @@ func TestGoWithTimeout(t *testing.T) {
 
 	require.Equal(t, uint32(testWorkerPoolSize+1), count)
 }
+
+func TestInitOnInitializedWorkerPool(t *testing.T) {
+	var count uint32
```
cw9: Feels a better test would be:
```
p := NewWorkerPool(testWorkerPoolSize)
	p.Init()

	var wg sync.WaitGroup
	wg.Add(1)
	go func() {
	  p.Init()
	  wg.Done()
	}
	wg.Wait()
```
This way it's clear that it's avoiding the hanging, rather than checking the counts and stuff

jskelcy:
@xichen2020 totally valid that we just use documentation to handle this issue. As you can see in the description this only came up from a fat finger and it struck me as kind of odd. I think either documentation, error, or noops are all valid solutions if we have a design principle that we stick to, but currently, we dont have any one of those solutions so we should pick one. 

@robskillington yep. I was trying to avoid a backwards breaking change, but you are right that does seem like the common pattern.

sync/worker_pool_test.go
jskelcy: Sure. I was thinking there should be some assert in the test but I suppose that is not necessarily true.

PR #145, opened 2018-04-24
Use strict YAML unmarshal when loading config

robskillington:
The current unmarshalling is susceptible to setting keys that are missing in the config resulting in unexpected default values when the key used is wrong.

This change makes config loading use the new `yaml.UnmarshalStrict` method now available in `gopkg.in/yaml.v2`.

### DIFF
--- a/config/config.go
+++ b/config/config.go
@@ -29,31 +29,41 @@ import (
 	yaml "gopkg.in/yaml.v2"
 )
 
-var errNoFilesToLoad = errors.New("attempt to load configuration with no files")
+var errNoFilesToLoad = errors.New("attempt to load config with no files")
 
-// LoadFile loads a config from a file.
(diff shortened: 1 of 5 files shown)


prateek:
Need to make the same linter changes as https://github.com/m3db/m3x/pull/143 for build to work (add `golint` to metalint config and remove lint target from `make all`)

config/config.go
```diff
@@ -50,7 +50,7 @@ func loadFiles(config interface{}, fnames ...string) error {
 			return err
 		}
 
-		if err := yaml.Unmarshal(data, config); err != nil {
+		if err := yaml.UnmarshalStrict(data, config); err != nil {
```
jeromefroe: Can we make this configurable with a default of on? For better or worse we depend on this in a few places already (e.g. collector).

PR #148, opened 2018-05-01
Leak Detecting Pool genny target

prateek:


generics/debugpool/pool.go
```diff
@@ -0,0 +1,129 @@
...
+// elemType is the generic type for use with the debug pool.
+type elemType generic.Type
+
+// elemTypeEqualsFn allows users to override equality checks
+// for `elemType` instances.
+type elemTypeEqualsFn func(a, b elemType) bool
+
+// elemTypePool is the backing pool wrapped by the debug pool.
+type elemTypePool interface {
+	generic.Type
```
robskillington: Hm why does this embed `interface{}`?  (generic.Type is just `interface{}` alias)

prateek: This is one of the nice bits of the forked genny version we're on. It's a way of forcing a type to conform to an interface. This decl gets stripped away during codegen.

prateek:
@robskillington ta, added the method and an example of using it.

PR #150, opened 2018-05-03
Add invariant violation metric

prateek:


instrument/invariant.go
```diff
@@ -0,0 +1,44 @@
...
+	InvariantViolatedLogPrefix = "[invariant violated]"
+)
+
+// EmitInvariantViolation emits a metric to indicate a system invariant has
+// been violated. Users of this method are expected to monitor/alert off this
+// metric to ensure they're notified when such an event occurs. Further, they
+// should log further information to aid diagnostics of the system invariant
+// violated at the callsite of the violation.
+func EmitInvariantViolation(opts Options) {
+	// NB(prateek): there's no need to cache this metric. It should be never
```
jeromefroe: I don't have much context here, but would it be preferable to just crash in the event of an invariant being violated and rely on the underlying process manager to restart the process rather than try to recover when certain assumptions may no longer be valid?

richardartoul: The motivation for this P.R came up in response to the data loss investigation I’ve been working on. As a result of that, we found several places in the codebase where we could add additional safeguards to detect issues.

A good number of these cases are bad enough to cause data loss, but if they are ever triggered they represent fundamental issues with M3DB that need to be addressed with bug fixes, and there would be no value in causing additional data loss by crashing the node. When stuff like this happens we emit logs, but we have no idea if these logs are being emitted on a regular basis unless we notice something strange and go look at them. With this metric we can have a low-priority alert so that we’re at least notified that the issue is actually manifesting in production.

In other situations where we’ve clearly reached a state that we should not be in and cannot proceed I agree we should just crash and restart.

prateek: Pretty much what Richie said. 

The choice of when to run in a degraded state v crashing needs to be made before using the utilities in this PR. This PR is to make a standard way to detect when we’ve chosen to run in a degraded state.

jeromefroe: Gotcha, I think my hangup was  just with the word invariant. The connotation I have when I hear invariant is that the property can never be violated. Any thoughts on calling it an assumption or something similar? Don’t feel too strongly here though

prateek: That's definitely intentional. This is going to used in places where we're doing the equivalent of an assert and can still continue processing (e.g. return error response to the user for a RPC code path that runs into an un-expected path). Going to retain the name to indicate severity.

instrument/invariant.go
```diff
@@ -0,0 +1,44 @@
...
+	// of an invariant.
+	InvariantViolatedLogPrefix = "[invariant violated]"
+)
+
+// EmitInvariantViolation emits a metric to indicate a system invariant has
+// been violated. Users of this method are expected to monitor/alert off this
+// metric to ensure they're notified when such an event occurs. Further, they
+// should log further information to aid diagnostics of the system invariant
+// violated at the callsite of the violation.
+func EmitInvariantViolation(opts Options) {
```
richardartoul: This all looks fine to me, my only question is do you think there are any places in the code-base where you might have access to a scope but not the InstrumentOptions? Probably unlikely right?

prateek: AFAIK all exported entry points into packages across our codebase pass around scopes using InstrumentOptions. Within a package we sometimes pass only the scope around but that's a minor refactor I don't mind expecting people to do if they want to use this method. Motivation is mainly because it feels weird to have a utility method on a scope live here, case Scope doesn't live here. At least if it operates on the InstrumentOptions, it kinda makes sense to keep here.

instrument/invariant_test.go
```diff
@@ -0,0 +1,40 @@
...
+)
+
+func ExampleInvariantViolatedMetricInvocation() {
+	testScope := tally.NewTestScope("", nil)
+	opts := instrument.NewOptions().SetMetricsScope(testScope)
+
+	instrument.EmitInvariantViolation(opts)
+	counters := testScope.Snapshot().Counters()
+	counter := counters[instrument.InvariantViolatedMetricName+"+"]
+	fmt.Println(counter.Name(), counter.Value())
```
robskillington: Shouldn't this be an assert equal to 1 and not a `fmt.Println`?

PR #156, opened 2018-05-04
Revert #154

prateek:
Found some transitive code that relies on the double alloc, going to revert for now.


prateek:
The bug was elsewhere, this code is kosher.

PR #157, opened 2018-05-05
Add reset capability to tag slice iterator so it can be reused

robskillington:


ident/tag_iterator.go
```diff
@@ -107,6 +105,17 @@ func (i *tagSliceIter) Duplicate() TagIterator {
 	}
 }
 
+func (i *tagSliceIter) Reset(tags Tags) {
+	i.backingSlice = tags
+	i.currentIdx = -1
+	if len(tags) != 0 {
+		// NB(r): To support calls to Current() before the first Next() is called.
```
prateek: Don't think you strictly need this. Calling `Current()` before `Next()` is a misuse of the iterator API, no?

PR #159, opened 2018-05-07
Add ident/testutil with new tags from tag iterator method

robskillington:


ident/testutil/tags_convert.go
```diff
@@ -0,0 +1,46 @@
...
+	var tags ident.Tags
+	if tagsLen := iter.Remaining(); tagsLen > 0 {
+		tags = make(ident.Tags, 0, tagsLen)
+		for iter.Next() {
+			curr := iter.Current()
+			tagName := checked.NewBytes(append([]byte(nil), curr.Name.Bytes()...), nil)
+			tagValue := checked.NewBytes(append([]byte(nil), curr.Value.Bytes()...), nil)
+			tags = append(tags, ident.BinaryTag(tagName, tagValue))
+		}
+		if err := iter.Err(); err != nil {
```
prateek: hm maybe do this check regardless of the length?

robskillington: Sure thing.

PR #161, opened 2018-05-11
Expose the math for calculating backoff time

cw9:
This pr exposes the function to return the backoff nanoseconds for a given retry, so I can reuse the math in a case where I don't actually need the attempt().

This pr changes the underlying math slightly.

@robskillington @xichen2020 @prateek @jeromefroe

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -124,7 +124,7 @@ func (r *retrier) attempt(continueFn ContinueFn, fn Fn) error {
 	r.metrics.errors.Inc(1)
 
 	for i := 0; r.forever || i < r.maxRetries; i++ {
-		r.sleepFn(time.Duration(r.BackoffNanos(i)))
+		r.sleepFn(time.Duration(BackoffNanos(i, r.jitter, r.backoffFactor, r.initialBackoff, r.maxBackoff)))
 
 		if continueFn != nil && !continueFn(attempt) {
 			return ErrWhileConditionFalse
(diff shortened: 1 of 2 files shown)


retry/retry.go
```diff
@@ -153,9 +146,30 @@ func (r *retrier) attempt(continueFn ContinueFn, fn Fn) error {
...
 		r.metrics.errors.Inc(1)
-		curr = int64(float64(curr) * r.backoffFactor)
 	}
 	r.metrics.errorsFinal.Inc(1)
 
 	return err
 }
+
+// BackoffNanos calculates the backoff for a retry in nanoseconds.
+func BackoffNanos(
```
jeromefroe: should this be private?

retry/retry.go
```diff
@@ -153,9 +146,30 @@ func (r *retrier) attempt(continueFn ContinueFn, fn Fn) error {
...
+	backoffFactor float64,
+	initialBackoff time.Duration,
+	maxBackoff time.Duration,
+) int64 {
+	backoff := initialBackoff.Nanoseconds()
+	if jitter {
+		half := backoff / 2
+		backoff = half + rand.Int63n(half)
+	}
+	if retry >= 1 {
```
jeromefroe: should this be before the jitter to match the previous behavior?

retry/retry.go
cw9: I need it to be public to be reused.

retry/retry.go
cw9: yeah will move this above, although it still does not give 100% the same math as before, but closer

retry/retry.go
jeromefroe: lol, yea I didn't read the PR message :)

PR #164, opened 2018-05-12
Add Tags type and NoFinalize() capability to ID, Tag and Tags

robskillington:
`Tags` is a new type that encapsulates a `[]Tag` slice and a pool so it can be passed around and pooled without having to hold a reference to the pool.  One can simply `Finalize()` the `Tags` type and have the slice be returned to the pool.

`NoFinalize()` is to be used when longly-lived types are being passed to sub-systems that have a mix of longly-lived types and shortly lived types and would like to uniformly finalize them all.  Components can return their types marked with `NoFinalize()` so that calling `Finalize()` on them is a no-op and can be safely returned without worry that the caller will impact the lifecycle of the type.


ident/types.go
```diff
@@ -33,10 +33,25 @@ import (
...
+	// Bytes returns the underlying byte slice of the bytes ID unpacked from
+	// any checked bytes container, callers cannot safely hold a ref to these
+	// bytes.
 	Bytes() []byte
+
+	// Equal returns whether the ID is equal to a given ID.
 	Equal(value ID) bool
-	Reset()
+
+	// NoFinalize makes calls to finalize a no-op, this is useful when you
```
prateek: +1

ident/types.go
```diff
@@ -48,12 +63,25 @@ type TagValue ID
...
+	Value      TagValue
+	noFinalize bool
 }
 
-// Finalize releases all resources held by the Tag.
+// NoFinalize makes calls to finalize a no-op, this is useful when you
+// would like to share a type with another sub-system that should is not
+// allowed to finalize the resource as the resource is kept indefinitely
+// until garbage collected (i.e. longly lived).
+func (t *Tag) NoFinalize() {
```
prateek: why not just call NoFinalize on t.Name and t.Value?

prateek: nvm i see why.

ident/types.go
```diff
@@ -103,17 +131,30 @@ type Pool interface {
...
+	PutTags(tags Tags)
+
 	// Clone replicates a given ID into a pooled ID.
 	Clone(id ID) ID
 
 	// CloneTag replicates a given Tag into a pooled Tag.
 	CloneTag(tag Tag) Tag
+
+	// CloneTags replicates a given set of Tags into a pooled Tags.
+	CloneTags(tags Tags) Tags
```
prateek: while you're here, maybe add `CloneTagIterator(...)` too?

ident/types.go
```diff
@@ -103,17 +131,30 @@ type Pool interface {
 	// string.
 	StringTag(name, value string) Tag
 
+	// GetTags will create a new array of tags and return it.  When the context
+	// closes the tags array and any tags contained will be finalized.
+	GetTags(ctx context.Context) Tags
```
prateek: same here, would be useful to have the ability to have an entry point which takes `ctx+tagIter` and gives back a Tags.

PR #165, opened 2018-05-14
Prevent panic in rand.Int63n

cw9:
@xichen2020

### DIFF
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -164,7 +164,7 @@ func BackoffNanos(
 	if retry >= 1 {
 		backoff = int64(float64(backoff) * math.Pow(backoffFactor, float64(retry-1)))
 	}
-	if jitter {
+	if jitter && backoff >= 2 {
 		half := backoff / 2
 		backoff = half + rand.Int63n(half)
 	}
--- a/retry/retry.go
+++ b/retry/retry.go
@@ -164,6 +164,7 @@ func BackoffNanos(
 	if retry >= 1 {
 		backoff = int64(float64(backoff) * math.Pow(backoffFactor, float64(retry-1)))
 	}
+	// Validate the value of backoff to make sure Int63n() does not panic.
 	if jitter && backoff >= 2 {
 		half := backoff / 2
 		backoff = half + rand.Int63n(half)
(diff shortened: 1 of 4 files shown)


retry/retry.go
```diff
@@ -164,7 +164,7 @@ func BackoffNanos(
 	if retry >= 1 {
 		backoff = int64(float64(backoff) * math.Pow(backoffFactor, float64(retry-1)))
 	}
-	if jitter {
+	if jitter && backoff >= 2 {
```
xichen2020: Can you add some clarifying comments please?

retry/retry.go
```diff
@@ -164,7 +164,8 @@ func BackoffNanos(
 	if retry >= 1 {
 		backoff = int64(float64(backoff) * math.Pow(backoffFactor, float64(retry-1)))
 	}
-	if jitter {
+	// Validate the value of backoff to make sure Int63n() does not panic.
+	if jitter && backoff >= 2 {
```
prateek: good to have this check but maybe add a `Validate()` to `Options` and ensure the user provided numbers are valid too.

retry/retry_test.go
```diff
@@ -193,3 +199,27 @@ func TestRetryForever(t *testing.T) {
 	assert.Equal(t, 10, numAttempts)
 	assert.Equal(t, time.Duration(1023*time.Second), totalSlept)
 }
+
+func TestBackoffNoPanic(t *testing.T) {
```
prateek: Nice!

retry/retry.go
cw9: I don't think the panic here is originated from invalid numbers in the options, the user could provide initialBackoff = 1 nano as a valid input and cause a panic in the original implementation.

prateek: That's what i mean. If we're going to silently not jitter unless `initialBackoff` >= 2, feels like we should inform the user that the value they gave us was wrong. Cause they'd expect the retrier to jitter cause they'd specified it should.

cw9: yeah even with initialBackoff == 1, say retry = 4, backoff will be  1 * 2^(4-1) = 8, which still can be jittered, so don't think initialBackoff is a blocker here, also that kind of value seems to be impractical in production, so I'd keep the interface simpler and save the validation call.

PR #177, opened 2018-06-04
Allow retrier to take in random source other than just the global one

cw9:


retry/types.go
```diff
@@ -39,6 +39,9 @@ func NonRetryableError(err error) error {
 	return errors.NewNonRetryableError(err)
 }
 
+// Int63nFn returns a non-negative pseudo-random number in [0,n).
+type Int63nFn func(n int64) int64
```
prateek: nit: `RngFn` or `RandInt64Fn` would be better names for this

cw9: Sounds good

PR #179, opened 2018-06-05
Implement pooled worker pool

richardartoul:
Tested on one of our production clusters and it reduced end-to-end latency significantly. In addition, the amount of time spent in runtime.morestack in the CPU profiles also decreased drastically.

### DIFF
--- a/sync/options.go
+++ b/sync/options.go
@@ -20,34 +20,42 @@
 
 package sync
 
+import "time"
+
 const (
-	defaultNumShards             = 2 ^ 4
+	defaultNumShards             = int64(2 ^ 4)
 	defaultKillWorkerProbability = 0.0001
-	defaultRandSeed              = int64(1337)
 )
+
+var (
+	defaultNowFn = time.Now
+)
+
+// NowFn is a function that returns the current time.
+type NowFn func() time.Time
 
 // NewPooledWorkerPoolOptions returns a new PooledWorkerPoolOptions with default options
 func NewPooledWorkerPoolOptions() PooledWorkerPoolOptions {
 	return &pooledWorkerPoolOptions{
 		numShards:             defaultNumShards,
 		killWorkerProbability: defaultKillWorkerProbability,
-		randSeed:              defaultRandSeed,
+		nowFn: defaultNowFn,
 	}
 }
 
 type pooledWorkerPoolOptions struct {
-	numShards             int
+	numShards             int64
 	killWorkerProbability float64
-	randSeed              int64
+	nowFn                 NowFn
 }
 
-func (o *pooledWorkerPoolOptions) SetNumShards(value int) PooledWorkerPoolOptions {
+func (o *pooledWorkerPoolOptions) SetNumShards(value int64) PooledWorkerPoolOptions {
 	opts := *o
 	opts.numShards = value
 	return &opts
 }
 
-func (o *pooledWorkerPoolOptions) NumShards() int {
+func (o *pooledWorkerPoolOptions) NumShards() int64 {
 	return o.numShards
 }
 
@@ -61,12 +69,12 @@ func (o *pooledWorkerPoolOptions) KillWorkerProbability() float64 {
 	return o.killWorkerProbability
 }
 
-func (o *pooledWorkerPoolOptions) SetRandSeed(value int64) PooledWorkerPoolOptions {
+func (o *pooledWorkerPoolOptions) SetNowFn(value NowFn) PooledWorkerPoolOptions {
 	opts := *o
-	opts.randSeed = value
+	opts.nowFn = value
 	return &opts
 }
 
-func (o *pooledWorkerPoolOptions) RandSeed() int64 {
-	return o.randSeed
+func (o *pooledWorkerPoolOptions) NowFn() NowFn {
+	return o.nowFn
 }
--- a/sync/options.go
+++ b/sync/options.go
@@ -20,52 +20,40 @@
 
 package sync
 
-import (
-	"sync"
-	"sync/atomic"
-	"testing"
-
-	"github.com/stretchr/testify/require"
+const (
+	defaultNumShards             = 10
+	defaultKillWorkerProbability = 0.0001
 )
 
-func TestPooledWorkerPoolGo(t *testing.T) {
-	var count uint32
-
-	p := NewPooledWorkerPool(testWorkerPoolSize, NewPooledWorkerPoolOptions())
-	p.Init()
-
-	var wg sync.WaitGroup
-	for i := 0; i < testWorkerPoolSize*2; i++ {
-		wg.Add(1)
-		p.Go(func() {
-			atomic.AddUint32(&count, 1)
-			wg.Done()
-		})
+// NewPooledWorkerPoolOptions returns a new PooledWorkerPoolOptions with default options
+func NewPooledWorkerPoolOptions() PooledWorkerPoolOptions {
+	return &pooledWorkerPoolOptions{
+		numShards:             defaultNumShards,
+		killWorkerProbability: defaultKillWorkerProbability,
 	}
-	wg.Wait()
-
-	require.Equal(t, uint32(testWorkerPoolSize*2), count)
 }
 
-func TestPooledWorkerPoolGoKillWorker(t *testing.T) {
-	var count uint32
+type pooledWorkerPoolOptions struct {
+	numShards             int
+	killWorkerProbability float64
+}
 
-	p := NewPooledWorkerPool(
-		testWorkerPoolSize,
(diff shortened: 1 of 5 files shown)


jskelcy:
Can you leave a quick comment about why you cant use the existing m3x worker pool implementation.

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,122 @@
...
+	numShards int
+}
+
+type workerInstruction struct {
+	work      Work
+	shouldDie bool
+}
+
+// NewPooledWorkerPool creates a new worker pool.
+func NewPooledWorkerPool(size int) PooledWorkerPool {
```
martin-mao: Nit: Add option for killProbability. Would be interesting to run in prod with a few different settings to prove out the effects in prod.

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,122 @@
...
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// Package sync implements synchronization facililites such as worker pools.
+package sync
+
+import "math/rand"
+
+const (
+	defaultNumShards = 10
```
martin-mao: Should we make this variable to optimize for HW with different cores?

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,122 @@
...
+type workerInstruction struct {
+	work      Work
+	shouldDie bool
+}
+
+// NewPooledWorkerPool creates a new worker pool.
+func NewPooledWorkerPool(size int) PooledWorkerPool {
+	numShards := defaultNumShards
+	if size < numShards {
+		numShards = size
```
martin-mao: Verify that size is not 0 otherwise numShards will be 0 and you'll panic on line 76

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,122 @@
...
+	for _, workCh := range p.workChs {
+		for i := 0; i < cap(workCh); i++ {
+			p.spawnWorker(workCh)
+		}
+	}
+}
+
+func (p *pooledWorkerPool) Go(work Work) {
+	instruction := workerInstruction{work: work}
+	randVal := rand.Float64()
```
martin-mao: Do you want to create a new rand and keep inside the pooledWorkerPool struct instead of using the globalRand?

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,122 @@
...
+	killWorker := randVal < defaultKillWorkerProbability
+	instruction.shouldDie = killWorker
```
martin-mao: Do you really need a struct with bool to be passed down the channel. Can you not do the rand inside the for loop inside the spawnWorker func? Then you can just have a chan of work, not workInstructions.

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,122 @@
...
+
+	// Pick an index based on the random value. Example:
+	// 		numShards = 10
+	// 		randVal = 0.14
+	// 		int(0.14 * float64(10)) = 1
+	// 		1 % 10 == 1
+	workChIdx := int((randVal * float64(p.numShards))) % p.numShards
```
martin-mao: Nit: Can you benchmark all this type casting VS generating a random int to use here. You can generate a random int to use here along with probability, just multiply your probability % option with the max int range you're going to use and check against that int.

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,122 @@
...
+	workCh := p.workChs[workChIdx]
+	workCh <- instruction
```
martin-mao: Consider using this as a non blocking channel and then inserting logic that if the channel happens to be full, move along the index (+1 % p.numSards) to find next empty work ch. We use a similar pattern in the cql driver. You will need to cap this at 1 full loop so they don't just iterate here forever, eg. after 1 full loop, you just block and wait.

sync/pooled_worker_pool.go
richardartoul: My original reasoning was that I did not want to blow the stack by creating goroutines inside of goroutines, but actually thats not how goroutines work at all. They don't become part of the spawning goroutines stack, so yeah I can fix this.

sync/pooled_worker_pool.go
richardartoul: Kind of annoying to have to return an error in this constructor though...

sync/pooled_worker_pool.go
richardartoul: made options

sync/pooled_worker_pool.go
richardartoul: made options

sync/pooled_worker_pool.go
richardartoul: although this way I only have to generate one random number

sync/pooled_worker_pool.go
richardartoul: I did some micro-benchmarks and there doesn't seem to be any (reliable) performance difference between getting a random float, and gettting a random float and then casting it to an int. We're talking about 20ns operations here so I don't think this will really impact anything compared to the amount of "work" being done. Plus I think its a little easier to follow this way.

sync/pooled_worker_pool.go
richardartoul: Ok turns out that getting a random number requires a lock so I opted to keep in the workerInstruction to prevent having to lock twice, but I moved the spawning logic into the workers themselves

sync/pooled_worker_pool.go
richardartoul: I don't think this will add much here. You should get a pretty good random distribution, and this complicates the logic a lot. It also means you're introducing a lot more contention when the pool is actually hot because instead of blocking and waiting you're doing all these random checks.

sync/pooled_worker_pool.go
richardartoul: done

sync/pooled_worker_pool.go
prateek: hm - how about using time.Now() as a pseudo rng instead of rand.Float64()? former should be cheaper considering you can avoid the lock contention yea?

sync/options.go
```diff
@@ -0,0 +1,72 @@
...
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+package sync
+
+const (
+	defaultNumShards             = 10
```
prateek: hm could you default to a power of 2 - 8/16?

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,107 @@
...
+	for _, workCh := range p.workChs {
+		for i := 0; i < cap(workCh); i++ {
+			p.spawnWorker(workCh)
+		}
+	}
+}
+
+func (p *pooledWorkerPool) Go(work Work) {
+	instruction := workerInstruction{work: work}
+	randVal := p.getRandFloat64()
```
prateek: hm how about breaking this into two pieces - use `time.Now() % numShards` to pick `workChIdx`. And have a new rand.Rand per spawnWorker? will avoid all the locking you need currently

richardartoul: I don't like the idea of using time.Now(), but I can just use different rand.Rand's for everything

sync/pooled_worker_pool.go
```diff
@@ -0,0 +1,88 @@
...
+	sync.Mutex
+	workChs               []chan Work
+	numShards             int64
+	killWorkerProbability float64
+	nowFn                 NowFn
+}
+
+// NewPooledWorkerPool creates a new worker pool.
+func NewPooledWorkerPool(size int, opts PooledWorkerPoolOptions) (PooledWorkerPool, error) {
+	if size <= 0 {
```
martin-mao: If you don't like the error return, you can always just use a defaultSize if they provide 0.

richardartoul: Yeah but its kind of sketch, better to be explicit. I just thought a panic was explicit enough, but this is good too :D

PR #181, opened 2018-06-06
Add multi worker pool

arnikola:


sync/multi_worker_pool.go
```diff
@@ -0,0 +1,65 @@
...
+	// GetPool provides a WorkerPool
+	GetPool() WorkerPool
+
+	// PutPool returns a WorkerPool to the pool
+	PutPool(pool WorkerPool)
+}
+
+type multiWorkerPool struct {
+	poolCh   chan WorkerPool
+	poolSize int
```
martin-mao: Nit: Call this inner pool size as not to confuse whether you're referring to how many pools are in the multi pool.

arnikola:
This may not be necessary, Object pool should be sufficient for what's required, and the cast isn't incurring too bad a cost.

arnikola:
Closing this as it's not necessary for the current applications

PR #185, opened 2018-07-25
 Add CurrentIndex and Len to ident.Iterator, ident.TagIterator for position

robskillington:


ident/iterator.go
```diff
@@ -54,6 +54,13 @@ func (i *idSliceIter) Current() ID {
 	return i.currentID
 }
 
+func (i *idSliceIter) CurrentIndex() int {
```
richardartoul: can it ever actually be less than zero?

PR #191, opened 2018-09-27
[WIP] Add a P-sharded Object Pool

prateek:


jeromefroe:
What do you think of following the example of [drwmutex](https://github.com/jonhoo/drwmutex) to make the pool per-CPU instead of per-P so we don't have to call into the runtime?

PR #193, opened 2018-09-28
Add GrowOnDemand option to pooled worker pool

richardartoul:


sync/pooled_worker_pool.go
```diff
@@ -73,14 +73,43 @@ func (p *pooledWorkerPool) Go(work Work) {
 	workCh <- work
 }
 
-func (p *pooledWorkerPool) spawnWorker(workCh chan Work) {
+func (p *pooledWorkerPool) GoOrGrow(work Work) {
+	// Use time.Now() to avoid excessive synchronization
+	workChIdx := p.nowFn().UnixNano() % p.numShards
+	workCh := p.workChs[workChIdx]
```
robskillington: Maybe move this code into a minor reused fn:
```
func (p *pooledWorkerPool) selectWorkCh() chan Work {
  workChIdx := p.nowFn().UnixNano() % p.numShards
  return p.workChs[workChIdx]
}
```

sync/pooled_worker_pool.go
```diff
@@ -73,14 +73,43 @@ func (p *pooledWorkerPool) Go(work Work) {
...
+		// allocate a new goroutine to do the work and then
+		// assign it to be a temporary additional worker for the queue.
+		// This allows the worker pool to accommodate "bursts" of
+		// traffic. Also, it reduces the need for operators to tune the size
+		// of the pool for a given workload. If the pool is initially
+		// sized too small, it will eventually grow to accommodate the
+		// workload, and if the workload decreases the killWorkerProbability
+		// will slowly shrink the pool back down to its original size because
+		// workers created in this manner will not spawn their replacement
+		// before killing themselves.
```
robskillington: Should we set a higher probability to kill these goroutines perhaps?

richardartoul: I think the issue with setting a higher probability is it limits how far the pool can grow. I guess that can be a good or bad thing depending on how you look at it, but I'd rather the pool take longer to shrink (as long as it does shrink) if it allows it to accommodate a wider variety of workloads

sync/pooled_worker_pool.go
richardartoul: sure

sync/types.go
```diff
@@ -45,6 +45,16 @@ type PooledWorkerPool interface {
...
+
+	// GoOrGrow tries to assign the work to an existing goroutine, but if
+	// none are available it will expand the pool of goroutines to accommodate
+	// the work. The newly allocated goroutine will temporarily participate
+	// in the pool in an effort to amortize its allocation cost, but will
+	// eventually be killed. This allows the pool to dynamically respond to
+	// workloads without causing excessive memory pressure. The pool will grow in
+	// size when the workload exceeds its capacity and shrink back down to its
+	// original size if/when the burst subsides.
+	GoOrGrow(work Work)
```
prateek: more an implementation nit: I'd prefer if we could keep the interface simple. At any single callsite, the user is always going to choose one of Go or GoOrGrow. So instead of exposing both methods, how about taking an option in the constructor to configures the behaviour of `Go()` to be either the current version or the bursty one.

Feels like a simpler API that way.

sync/pooled_worker_pool.go
prateek: Hm don't have to change the probability for all go routines, only the new ones created during the burst. That should work yea?

sync/pooled_worker_pool.go
```diff
@@ -61,26 +61,57 @@ func NewPooledWorkerPool(size int, opts PooledWorkerPoolOptions) (PooledWorkerPo
...
+	workCh := p.selectWorkCh()
+	workCh <- work
+}
+
+func (p *pooledWorkerPool) GoOrGrow(work Work) {
+	workCh := p.selectWorkCh()
+	select {
+	case workCh <- work:
+	default:
+		// If the queue for the worker we were assigned to is full,
```
prateek: maybe add some (sampled) instrumentation to make this behaviour observable

PR #200, opened 2018-10-05
Increase default kill probability for dynamic worker pools

arnikola:


config/pooling.go
```diff
@@ -24,6 +24,7 @@ import "github.com/m3db/m3x/sync"
 
 const (
 	defaultWorkerPoolStaticSize = 4096
+	defaultGrowKillProbability  = 0.1
```
richardartoul: 0.01 (1% not 10%)

arnikola: Oops, yeah this may be a bit high haha

PR #204, opened 2018-11-22
[hostid] support reading ID from a file

schallert:
Under some orchestration systems, such as Kubernetes, we may need to
read host identity from a file on disk that is dynamically populated
with instance metadata. Additionally, this metadata may sometimes be
populated by an external actor and may not be available at startup, so
we provide an optional timeout to wait for it.

As a concrete example, on Kubernetes this supports reading host ID from
pod metadata that is injected using [downward API
volumes](https://git.io/fp4hL). The host ID may not be immediately
populated, but the user can provide a timeout that the instance will
wait for it before failing to start. An example of a pod metadata file
in Kubernetes is:

```
root@matt-test-656988ff69-gk49b:/# cat /etc/podinfo/id
test-id
```

### DIFF
--- a/config/hostid/hostid.go
+++ b/config/hostid/hostid.go
@@ -27,11 +27,16 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
+	"strings"
 	"time"
 )
 
 const (
-	defaultFileCheckInterval = 10 * time.Second
+	defaultFileCheckInterval = time.Second
+)
+
+var (
+	errHostIDFileEmpty = errors.New("host ID file is empty")
 )
 
 // Resolver is a type of host ID resolver
@@ -65,17 +70,17 @@ type Configuration struct {
 	// EnvVarName is the environment specified host ID if using environment host ID resolver.
 	EnvVarName *string `yaml:"envVarName"`
 
-	// File is the kv file config.
+	// File is the file config.
 	File *FileConfig `yaml:"file"`
 }
 
 // FileConfig contains the info needed to construct a FileResolver.
 type FileConfig struct {
-	// Path of the file containing the KV config.
+	// Path of the file containing the host ID.
 	Path string `yaml:"path"`
 
 	// Timeout to wait for the file to be non-empty.
-	Timeout time.Duration `yaml:"timeout"`
+	Timeout *time.Duration `yaml:"timeout"`
 }
 
 func (c Configuration) resolver() (IDResolver, error) {
@@ -88,7 +93,7 @@ func (c Configuration) resolver() (IDResolver, error) {
 		return &environmentResolver{envVarName: c.EnvVarName}, nil
 	case FileResolver:
 		if c.File == nil {
-			return nil, errors.New("KV file config cannot be nil")
+			return nil, errors.New("file config cannot be nil")
 		}
 		return &file{
 			path:    c.File.Path,
@@ -148,7 +153,7 @@ func (c *environmentResolver) ID() (string, error) {
 type file struct {
 	path     string
 	interval time.Duration
-	timeout  time.Duration
+	timeout  *time.Duration
 }
 
 // ID attempts to parse an ID from a  file. It will optionally wait a timeout to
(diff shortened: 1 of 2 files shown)


config/hostid/hostid.go
```diff
@@ -50,34 +67,151 @@ type Configuration struct {
...
+			return v, nil
+		}
+
+		time.Sleep(interval)
+	}
+
+	return "", fmt.Errorf("did not find value in %s within %s", c.path, c.timeout)
+}
+
+func (c *kvFile) parseKV(r io.Reader) (string, error) {
```
robskillington: Any reason not to use JSON or YAML for this?  Seems a little strange to implement parser yourself.  

You could also just make it a plain text file, with nothing else in it. (so `ioutil.ReadAll(...)`)

robskillington:
Just wondering if alternatively you could just avoid launching the process until the file is there?  That way this can be entirely out of process, and then the env variable just set.

i.e.
```bash
#!/bin/bash
# m3db_start.sh

# Wait for host ID
HOST_ID_VALUE=""
while [ "$HOST_ID_VALUE" = "" ]; do
  # check here if slept too long in total... and exit 0 if so

  if ![ -f /my/file ]; then
    sleep 1
  fi

  HOST_ID_VALUE="$(cat /my/file)"
done

# Launch M3DB
HOST_ID="$HOST_ID_VALUE" ./m3db -f ...
```

config/hostid/hostid.go
schallert: Unfortunately we're limited to the format that Kubernetes spits out the annotations in, but I just realized that we can configure just 1 annotation field to be dumped and then could read the file's content as you mentioned. Will update.

schallert:
> Just wondering if alternatively you could just avoid launching the process until the file is there? That way this can be entirely out of process, and then the env variable just set.

I might be able to do this by having an [init container](https://v1-11.docs.kubernetes.io/docs/concepts/workloads/pods/init-containers/) that waits until the file has content before allowing the m3db pod to start up. I'll test that out and update this PR as necessary.

config/hostid/hostid.go
```diff
@@ -23,8 +23,15 @@
...
 import (
+	"errors"
 	"fmt"
+	"io/ioutil"
 	"os"
+	"time"
+)
+
+const (
+	defaultFileCheckInterval = 10 * time.Second
```
robskillington: Let's make this 1s, its not expensive at all to stat a file?

config/hostid/hostid.go
```diff
@@ -50,34 +64,134 @@ type Configuration struct {
...
-		return *c.Value, nil
+		return &configResolver{value: c.Value}, nil
 	case EnvironmentResolver:
-		if c.EnvVarName == nil {
-			err := fmt.Errorf("missing host ID env var name using: resolver=%s",
-				string(c.Resolver))
+		return &environmentResolver{envVarName: c.EnvVarName}, nil
+	case FileResolver:
+		if c.File == nil {
+			return nil, errors.New("KV file config cannot be nil")
```
robskillington: nit: Remove "KV" from this now?

config/hostid/hostid.go
```diff
@@ -50,34 +64,134 @@ type Configuration struct {
...
-			err := fmt.Errorf("missing host ID env var value using: resolver=%s, name=%s",
-				string(c.Resolver), *c.EnvVarName)
+
+		data, err := ioutil.ReadAll(f)
+		if err != nil {
 			return "", err
 		}
-		return v, nil
+
+		if len(data) == 0 {
```
robskillington: I would do the following, to help get rid of newlines, etc:
```go
value := strings.TrimSpace(string(data))
if len(value) == 0 {
  return "", err...
}
return value, nil
```

config/hostid/hostid.go
```diff
@@ -50,34 +64,134 @@ type Configuration struct {
...
+			return "", errors.New("host ID cannot be empty")
```
robskillington: nit: Make this a `errFileHostIDEmpty = errors.New(...)`?

config/hostid/hostid.go
```diff
@@ -50,34 +64,134 @@ type Configuration struct {
...
 
-// Resolve returns the resolved host ID given the configuration.
-func (c Configuration) Resolve() (string, error) {
+// FileConfig contains the info needed to construct a FileResolver.
+type FileConfig struct {
+	// Path of the file containing the KV config.
+	Path string `yaml:"path"`
+
+	// Timeout to wait for the file to be non-empty.
+	Timeout time.Duration `yaml:"timeout"`
```
robskillington: nit: Maybe make this a pointer type and use a default Timeout if not specified?  Of say 10s?

PR #209, opened 2019-01-25
Add tracing config

andrewmains12:
This is a small preparatory diff to add support for tracing to m3query. I figured I might as well do the config here, since that's where we have metrics + logging. Let me know if you'd rather have this code elsewhere!

### DIFF
--- a/glide.yaml
+++ b/glide.yaml
@@ -47,6 +47,14 @@ import:
   subpackages:
   - cmp
 
+- package: github.com/uber/jaeger-lib
+  version: ^1.5.0
+
+- package: github.com/uber/jaeger-client-go
+  version: ^2.7.0
+
+
+
 testImport:
 - package: github.com/fortytw2/leaktest
   version: ^1.1.0
--- a/glide.yaml
+++ b/glide.yaml
@@ -53,8 +53,6 @@ import:
 - package: github.com/uber/jaeger-client-go
   version: ^2.7.0
 
-
-
 testImport:
 - package: github.com/fortytw2/leaktest
   version: ^1.1.0
--- a/instrument/tracing.go
+++ b/instrument/tracing.go
@@ -1,0 +1,54 @@
+package instrument
+
+import (
+	"fmt"
+	"io"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/uber-go/tally"
+	jaegercfg "github.com/uber/jaeger-client-go/config"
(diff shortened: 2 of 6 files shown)


glide.yaml
```diff
@@ -47,6 +47,14 @@ import:
...
   - cmp
 
+- package: github.com/uber/jaeger-lib
+  version: ^1.5.0
+
+- package: github.com/uber/jaeger-client-go
+  version: ^2.7.0
+
+
+
```
benraskin92: Remove some of these lines?

instrument/tracing.go
```diff
@@ -0,0 +1,54 @@
...
+
+	cfg.Jaeger.ServiceName = serviceName
+	jaegerLog := jaegerzap.NewLogger(logger)
+	tracer, jaegerCloser, err := cfg.Jaeger.NewTracer(
+		jaegercfg.Logger(jaegerLog),
+		jaegercfg.Metrics(jaegertally.Wrap(scope)))
+
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to initialize jaeger: %s", err.Error())
+	}
```
benraskin92: newline

instrument/tracing.go
```diff
@@ -0,0 +1,54 @@
...
+}
+
+
+func (cfg *TracingConfiguration) NewTracer(serviceName string, scope tally.Scope, logger *zap.Logger) (opentracing.Tracer, io.Closer, error) {
+	if cfg.Backend == "" {
+		return opentracing.NoopTracer{}, noopCloser{}, nil
+	}
+
+	if cfg.Backend != TracingBackendJaeger {
+		return nil, nil, fmt.Errorf("unknown tracing backend: %s", cfg.Backend)
```
benraskin92: Maybe add that `jaeger` is the only supported backend?

instrument/tracing.go
```diff
@@ -0,0 +1,54 @@
...
+	}
+
+	cfg.Jaeger.ServiceName = serviceName
```
benraskin92: Check to make sure `cfg.Jaeger.ServiceName != ""`

glide.yaml
andrewmains12: Whoops yeah will do.

instrument/tracing.go
andrewmains12: Good call!

instrument/tracing.go
andrewmains12: Good call.

PR #211, opened 2019-02-12
Use runtime.NumCPU() for num shards and kill worker probability to 0.001

richardartoul:


sync/options.go
```diff
@@ -21,18 +21,19 @@
...
+	"runtime"
 	"time"
 
 	"github.com/m3db/m3x/instrument"
 )
 
 const (
-	defaultNumShards             = int64(2 ^ 4)
-	defaultKillWorkerProbability = 0.0001
+	defaultKillWorkerProbability = 0.001
```
arnikola: Do we need to edit the default grow kill probability in `config/pooling.go`?

sync/options.go
```diff
@@ -21,18 +21,19 @@
...
 	defaultGrowOnDemand          = false
 )
 
 var (
+	defaultNumShards int64 // Set by init().
```
justinjc: Why not just initialize this to `int64(runtime.NumCPU())` here?

richardartoul: for some reason I thought I couldn't call a function, will fix

sync/options.go
richardartoul: good call, done

PR #212, opened 2019-03-27
Add goCtx to xCtx

benraskin92:
Add Go's stdlib context to m3x's context

### DIFF
--- a/context/context.go
+++ b/context/context.go
@@ -29,6 +29,7 @@ import (
 )
 
 // NB(r): using golang.org/x/net/context is too GC expensive.
+// Instead, we just embed one.
 type ctx struct {
 	sync.RWMutex
 
--- a/context/context.go
+++ b/context/context.go
@@ -21,11 +21,10 @@
 package context
 
 import (
+	stdctx "context"
 	"sync"
 
 	"github.com/m3db/m3x/resource"
-
-	xnetcontext "golang.org/x/net/context"
 )
 
 // NB(r): using golang.org/x/net/context is too GC expensive.
@@ -33,7 +32,7 @@ import (
 type ctx struct {
 	sync.RWMutex
 
-	goCtx         xnetcontext.Context
+	goCtx         stdctx.Context
 	pool          contextPool
 	done          bool
 	wg            sync.WaitGroup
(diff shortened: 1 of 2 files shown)


context/context.go
```diff
@@ -24,12 +24,16 @@ import (
 	"sync"
 
 	"github.com/m3db/m3x/resource"
+
+	xnetcontext "golang.org/x/net/context"
```
jeromefroe: Why not use the [standard library package](https://golang.org/pkg/context/)?

benraskin92: Wanted to be consistent with what we use in m3db

jeromefroe: They are type aliases for one another, you can use the std lib and be consistent.

jeromefroe:
```go
package context

import "context" // standard library's context, as of Go 1.7

// A Context carries a deadline, a cancelation signal, and other values across
// API boundaries.
//
// Context's methods may be called by multiple goroutines simultaneously.
type Context = context.Context
```

context/context.go
```diff
@@ -51,6 +54,18 @@ func newPooledContext(pool contextPool) Context {
 	return &ctx{pool: pool}
 }
 
+func (c *ctx) GoContext() (xnetcontext.Context, bool) {
+	if c.goCtx == nil {
+		return nil, false
+	}
+
+	return c.goCtx, true
```
arnikola: `return c.goCtx, c.goCtx != nil`? Maybe that's getting too cute though

benraskin92: too cute

context/context.go
```diff
@@ -51,6 +54,18 @@ func newPooledContext(pool contextPool) Context {
 	return &ctx{pool: pool}
 }
 
+func (c *ctx) GoContext() (xnetcontext.Context, bool) {
```
arnikola: Should this be called `GoContext` if it's `golang.org/x/net/context`? I'd expect GoContext to use `context`

benraskin92: As Jerome pointed out `golang.org/x/net/context` is just an alias for `context`. Either way, I changed it to just use `context`.

PR #214, opened 2019-03-31
Move opentracing util to m3x

benraskin92:


opentracing/tracing_test.go
```diff
@@ -18,7 +18,7 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-package instrument
+package opentracingutil
```
prateek: nit: rename so as to keep the package name the same as the folder name.

opentracing/context.go
```diff
@@ -0,0 +1,80 @@
...
+
+package opentracingutil
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/opentracing/opentracing-go"
+	"github.com/opentracing/opentracing-go/log"
+	"golang.org/x/net/context"
```
prateek: any reason to prefer this over the standard lib "context" here?

opentracing/context.go
```diff
@@ -0,0 +1,80 @@
...
+	} else {
+		tracer = getGlobalTracer()
+	}
+
+	span := tracer.StartSpan(operationName, opts...)
+
+	return span, opentracing.ContextWithSpan(ctx, span)
+}
+
+// Time is a log.Field for time.Time values. It translates to RF3339 formatted time strings.
```
prateek: nit: `RFC3339`. Maybe also add an example of what that means, e.g. `"2019-04-01T15:04:05Z07:00"`

opentracing/context.go
benraskin92: Nope  - I'll switch.

PR #216, opened 2019-04-03
[context] Add child ctx

benraskin92:


context/context.go
```diff
@@ -131,10 +159,30 @@ const (
 )
 
 func (c *ctx) Close() {
+	parent := c.ParentCtx()
+	if parent != nil {
+		c.returnToPool()
```
robskillington: nit: Return to pool after calling close on parent.

context/context.go
```diff
@@ -131,10 +159,30 @@ const (
...
+		return
+	}
+
 	c.close(closeAsync)
 }
 
 func (c *ctx) BlockingClose() {
+	parent := c.ParentCtx()
+	if parent != nil {
+		c.returnToPool()
```
robskillington: nit: Return to pool after calling close on parent.

context/context.go
```diff
@@ -202,3 +256,60 @@ func (c *ctx) returnToPool() {
...
+	} else {
+		child = c.pool.Get()
+	}
+
+	child.SetParentCtx(c)
+
+	return child
+}
+
+func (c *ctx) SetParentCtx(parentCtx Context) {
```
robskillington: Probably don't want to make this method exported, just have an internal setter.

context/context.go
```diff
@@ -202,3 +256,60 @@ func (c *ctx) returnToPool() {
...
+	// check to see if parent exists?
+	c.parent = parentCtx
+}
+
+func (c *ctx) ParentCtx() Context {
```
robskillington: I think you should pick one of:
```
NewChildContext()
ParentContext()
```

or
```
NewChild()
Parent()
```

For consistency.  Also I would optimize the locking here to avoid defer.

context/context.go
```diff
@@ -202,3 +256,60 @@ func (c *ctx) returnToPool() {
...
+		return c, nil, false
+	}
+
+	var (
+		sp    opentracing.Span
+		spCtx stdctx.Context
+	)
+
+	sp = opentracing.SpanFromContext(goCtx)
+	jaegerSpan, ok := sp.(*jaeger.Span)
```
robskillington: Put a note that until `opentracing` supports the `IsSampled()` method we need to cast to jaeger span.  

Maybe we can link to an open issue too? Seems there are two issues (one is a duplicate):
https://github.com/opentracing/specification/issues/8
https://github.com/opentracing/specification/issues/92

context/context_test.go
```diff
@@ -105,7 +161,7 @@ func TestDependsOnNoCloserAllocation(t *testing.T) {
 	assert.Nil(t, ctx.finalizeables)
 }
 
-func TestDependsOn(t *testing.T) {
+func TestDependsOnX(t *testing.T) {
```
robskillington: Did you mean to change this method name?  Probably can remove the "X".

benraskin92: Will revert back - was just testing stuff.

context/types.go
```diff
@@ -71,6 +71,12 @@ type Context interface {
 
 	// SetGoContext sets the Go std context
 	SetGoContext(stdctx.Context)
+
+	// ParentCtx returns the parent context
+	ParentCtx() Context
+
+	// SetParentCtx sets the parent context
+	SetParentCtx(Context)
```
robskillington: Yeah I would remove these methods altogether and just add the following:
```go
StartSpan(name string) (Context, opentracing.Span, bool)
```

context/context.go
robskillington: Actually on second thoughts, let's just unexport all these methods - we only need to add StartSpan for now.

context/context.go
```diff
@@ -202,3 +256,60 @@ func (c *ctx) returnToPool() {
...
+	}
+
+	child.SetParentCtx(c)
+
+	return child
+}
+
+func (c *ctx) SetParentCtx(parentCtx Context) {
+	// check to see if parent exists?
+	c.parent = parentCtx
```
robskillington: Need to lock here.

context/context.go
```diff
@@ -202,3 +256,60 @@ func (c *ctx) returnToPool() {
...
+}
+
+func (c *ctx) ParentCtx() Context {
+	c.RLock()
+	defer c.RUnlock()
+	return c.parent
+}
+
+func (c *ctx) StartSpan(name string) (Context, opentracing.Span, bool) {
```
robskillington: Can we rename this to `StartTraceSpan` since then the usage is clear it's for a trace?

context/context.go
```diff
@@ -202,3 +255,82 @@ func (c *ctx) returnToPool() {
...
+
+func (c *ctx) newChildContext() Context {
+	var child Context
+	if c.pool == nil {
+		child = NewContext()
+	} else {
+		child = c.pool.Get()
+	}
+
+	// is there a better way to do this?
```
robskillington: Hm, maybe try to cast it and if it fails then create one yourself?  Maybe we should create the following two methods so there is always one way to get a concrete `ctx`.

i.e. something like this:
```go

func NewContext() Context {
  return newContext()
}

func newContext() *ctx {
  return &ctx{
    //.. whatever
  }
}

// ..

func (c *ctx) newChild() Context {
	var child *ctx
	if c.pool != nil {
		pooled, ok := c.pool.Get().(*ctx)
		if ok {
			child = pooled
		}
	}

	if child == nil {
		child = newContext() // which returns *ctx
	}

	child.setParentCtx(c)
	return child
}

```

context/context.go
robskillington: Need to add this note above this method still FYI.

context/context.go
```diff
@@ -60,6 +60,11 @@ func newPooledContext(pool contextPool) Context {
 	return &ctx{pool: pool}
 }
 
+// newContext returns an empty ctx
+func newContext() *ctx {
```
robskillington: nit: Can you change `NewContext` to call `newContext()` instead of `newPooledContext(nil)`?

